PR #40, opened 2019-10-10
Update to build/command functions to write to slave controllers.

julesfowler:
So, initial thoughts on adding functionality to access the slave controllers.
As documented in the JIRA ticket, to access the "master" controller, command is the same. Example : `1PR50`, to move the first motor on the master controller 50 steps.
To move a slave controller, you prepend that to the command.
Example : `2>1PR50`, to move the first linked controller's first motor 50 steps.

TO DO :
The two ticked boxes are *vital* for moving forward with the picomotors, the rest are just good to haves -- so I'm going to un-draft this and ask for review. 

- [x] Write daisy chained controller access into `__build_message` and `command`.
- [ ] Find a smart way of exploring how many controllers are plugged in. (For the programmatic resets of positions to zero, it's easy to loop through all 4 motors but it's silly to loop through up to 31 slave controllers just in case.)
- [x] Update the `reset_controller` function such that we go through all controllers all motors.
- [ ] Annnnd, if we've figured out how to explore the controller # programmatically, we've probably also figured out how to get how many motors are plugged in, so might as well fold that into clever error handling for this.


catkit/interfaces2/newport_picomotor.py
```diff
@@ -282,8 +291,12 @@ def _build_message(self, cmd_key, cmd_type, axis=None, value=None):
...
             The message to send to the controller site.
         """
-        
-        address = self.cmd_dict[cmd_key]
+        if daisy_key is None or isinstance(daisy_key, int):
+            daisy_prefix = "{}>".format(daisy_key) if daisy_key else ""
+        else:
+            raise TypeError("The 'daisy_key' requires an integer or None input.")
+
+        address = daisy_prefix + self.cmd_dict[cmd_key]
```
jamienoss, line 308: Ok, so this is the _only_ aspect of the "daisy key", correct?

I would remove the semantics of daisy key, and just use something like ``controller_id``. Remove this from _all_ func sigs other than the ``__init__()``, in which it is a keyword arg with the default for whatever a non-daisy-chained controller is, which from your code looks like ``None``.

jamienoss: The check needs to be moved to the ``__init__()``.
Leaving this part of the code to be:
```python
address = f"{self.controller_id}>{self.cmd_dict[cmd_key]}" if self.controller_id else self.cmd_dict[cmd_key]
```

catkit/interfaces2/newport_picomotor.py
```diff
@@ -58,6 +62,10 @@ def __init__(self, ip=None, max_step=None, timeout=None, home_reset=True):
             Whether or not to reset to the home position on controller close.
             Defaults to True.
         """
+        
+        # Build a list of motors and controllers. 
+        self.motors = [n for n in range(1, motor_count+1)]
+        self.controllers = [None] + [n for n in range(2, controller_count+1)]
```
jamienoss, line 68: For future ref, it would be better to not assume that the motors be contiguously connected as ``motor_count`` does. Instead, passing in a list of motors would be just as trivial and more explicit.

catkit/interfaces2/newport_picomotor.py
```diff
@@ -58,6 +62,10 @@ def __init__(self, ip=None, max_step=None, timeout=None, home_reset=True):
             Whether or not to reset to the home position on controller close.
             Defaults to True.
         """
+        
+        # Build a list of motors and controllers. 
+        self.motors = [n for n in range(1, motor_count+1)]
```
jamienoss, line 67: Also, ``list(range(1, motor_count+1))``

PR #84, opened 2020-04-16
HiCAT-687: implement phase retrieval code on HiCAT

grbrady:
Added code in DmCommand to create a flat map with constant voltages or to create a flat map with voltages perturbed by Zernikes.

These are necessary to support determining new flat maps using phase retrieval data in the [HiCAT-687 branch of hicat-package](https://github.com/spacetelescope/hicat-package/tree/HiCAT-687-implement-phase-retrieval-code-on-HiCAT).

Upstream PR for https://github.com/spacetelescope/hicat-package/pull/248

### DIFF
--- a/catkit/hardware/boston/DmCommand.py
+++ b/catkit/hardware/boston/DmCommand.py
@@ -355,10 +355,14 @@ def add_zernike_to_flat_map(output_path, file_name=None, dm_num=1, zernike_index
     flat_map_volts = get_flat_map_volts(dm_num)
 
     flat_map_volts += (hicat_zernike_module.create_zernike(zernike_index, zernike_coeff_volts) + zernike_coeff_volts/2)
-    flat_map_volts *= catkit.util.get_dm_mask()
+    mask = catkit.util.get_dm_mask().astype('bool')
+    flat_map_volts *= mask
+    flat_map_volts_masked = flat_map_volts[mask]
+    flat_map_volts_masked[flat_map_volts_masked < 0] = 0
+    flat_map_volts[mask] = flat_map_volts_masked
 
     if file_name is None:
-        file_name = f"flat_map_volts_dm_{dm_num}_zernike_{zernike_index}_{zernike_coeff_volts}V.fits"
+        file_name = f"flat_map_volts_dm_{dm_num}_zernike_{zernike_index}_{zernike_coeff_volts:.2f}V.fits"
 
     catkit.util.write_fits(flat_map_volts, os.path.join(output_path, file_name))
 
--- a/catkit/hardware/boston/DmCommand.py
+++ b/catkit/hardware/boston/DmCommand.py
@@ -288,7 +288,6 @@ def convert_dm_image_to_command(dm_image, path_to_save=None):
         catkit.util.write_fits(dm_command, path_to_save)
     return dm_command
 
-
 def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None, dm_num=1):
     """
     Converts a dm_command_2d.fits to the format used for the flatmap, and outputs a new flatmap fits file.
@@ -313,4 +312,27 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
     # Convert the dm command units to volts.
     max_volts = CONFIG_INI.getint("boston_kilo952", "max_volts")
     dm_command_data *= max_volts
-    catkit.util.write_fits(dm_command_data, output_path)
+    catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
+
+
+def create_constant_flat_map(output_path, file_name=None, dm_num=1):
+    """
+    Creates a uniform flat map and outputs a new flatmap fits file.
+    :param output_path: Path to output the new flatmap fits file. Default is hardware/boston/
+    :param file_name: Filename for new flatmap fits file. Default is
+            flat_map_volts_dm_<1 or 2>_constant.fits
+    ;param dm_num: Which DM is this for?  Defaults to 1.
+    :return: None
+    """
+    if file_name is None:
+        # Create a string representation of the current timestamp.
+        # time_stamp = time.time()
+        # date_time_string = datetime.datetime.fromtimestamp(time_stamp).strftime("%Y-%m-%dT%H-%M-%S")
+        file_name = "flat_map_volts_dm_" + str(dm_num) + "_constant.fits"
+
+    bias_volts = CONFIG_INI.getint('boston_kilo952', f'bias_volts_dm{dm_num}')
+    mask = catkit.util.get_dm_mask()
+    dm_command_data = mask * bias_volts
+    catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
+
+
--- a/catkit/hardware/boston/DmCommand.py
+++ b/catkit/hardware/boston/DmCommand.py
@@ -111,11 +111,12 @@ class DmCommand(object):
                     flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm1")
                     flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
                     dm_command += flat_map_volts[0].data
-                    # dm_command -= 90
+                    # dm_command -= 30
                 else:
                     flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm2")
                     flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
                     dm_command += flat_map_volts[0].data
+                    # dm_command -= 30
 
             # Convert between 0-1.
             dm_command /= self.max_volts
@@ -336,4 +337,26 @@ def create_constant_flat_map(output_path, file_name=None, dm_num=1):
     dm_command_data = mask * bias_volts
     catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
 
+def add_zernike_to_flat_map(output_path, file_name=None, dm_num=1):
+    """
+    Creates a uniform flat map and outputs a new flatmap fits file.
+    :param output_path: Path to output the new flatmap fits file. Default is hardware/boston/
+    :param file_name: Filename for new flatmap fits file. Default is
+            flat_map_volts_dm_<1 or 2>_constant.fits
+    ;param dm_num: Which DM is this for?  Defaults to 1.
+    :return: None
+    """
+    flat_map = get_flat_map_volts(dm_num)
+
+
+    if file_name is None:
+        # Create a string representation of the current timestamp.
+        # time_stamp = time.time()
+        # date_time_string = datetime.datetime.fromtimestamp(time_stamp).strftime("%Y-%m-%dT%H-%M-%S")
+        file_name = "flat_map_volts_dm_" + str(dm_num) + "_constant.fits"
+
+    bias_volts = CONFIG_INI.getint('boston_kilo952', f'bias_volts_dm{dm_num}')
+    mask = catkit.util.get_dm_mask()
+    dm_command_data = mask * bias_volts
+    catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
 
--- a/catkit/hardware/boston/DmCommand.py
+++ b/catkit/hardware/boston/DmCommand.py
@@ -339,7 +339,7 @@ def create_constant_flat_map(output_path, file_name=None, dm_num=1):
     dm_command_data = mask * bias_volts
     catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
 
-def add_zernike_to_flat_map(output_path, file_name=None, dm_num=1, zernike_index=4, zerike_coeff_volts=10):
+def add_zernike_to_flat_map(output_path, file_name=None, dm_num=1, zernike_index=4, zernike_coeff_volts=10):
     """
     Creates a new flat map from the existing flat map for selected DM by adding a specified Zernike term with the
     specified coefficient magnitude.
@@ -354,8 +354,8 @@ def add_zernike_to_flat_map(output_path, file_name=None, dm_num=1, zernike_index
     """
     flat_map_volts = get_flat_map_volts(dm_num)
 
-    flat_map_volts += hicat_zernike_module.create_zernike(zernike_index, zerike_coeff_volts)
-    flat_map_volts *= mask
+    flat_map_volts += (hicat_zernike_module.create_zernike(zernike_index, zernike_coeff_volts) + zernike_coeff_volts/2)
+    flat_map_volts *= catkit.util.get_dm_mask()
 
     if file_name is None:
         file_name = f"flat_map_volts_dm_{dm_num}_zernike_{zernike_index}_{zernike_coeff_volts}V.fits"
--- a/catkit/hardware/boston/DmCommand.py
+++ b/catkit/hardware/boston/DmCommand.py
@@ -14,6 +14,9 @@ m_per_volt_map2 = None  # for caching the conversion factor, to avoid reading fr
 
 
 class DmCommand(object):
+
+    valid_dm_nums = (1, 2)
+
     def __init__(self, data, dm_num, flat_map=False, bias=False, as_voltage_percentage=False,
                  as_volts=False, sin_specification=None):
         """
@@ -26,7 +29,8 @@ class DmCommand(object):
         :param dm_num: Valid values are 0, 1, 2.  Default is 0, which will automatically identify which dm to use.
                      You can specify 1 or 2 to ensure a certain DM is used, regardless of how the input data is padded.
         :param flat_map: If true, add flat map correction to the data before outputting commands
-        :param bias: If true, add bias to the data before outputting commands
+        :param bias: None|bool|int
+            The bias voltage to add to the data before outputting commands. If None|True values are pulled from config.
         :param as_voltage_percentage: Interpret the data as a voltage percentage instead of meters; Deprecated.
         :param as_volts: If true, interpret the data as volts instead of meters
         :param sin_specification: Add this sine to the data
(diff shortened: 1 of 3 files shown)


catkit/hardware/boston/DmCommand.py
```diff
@@ -111,10 +113,12 @@ def to_dm_command(self):
                     flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm1")
                     flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
                     dm_command += flat_map_volts[0].data
+                    #dm_command -= 15
```
ivalaginja: If you don't need this line, please delete

catkit/hardware/boston/DmCommand.py
```diff
@@ -111,10 +113,12 @@ def to_dm_command(self):
...
                 else:
                     flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm2")
                     flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
                     dm_command += flat_map_volts[0].data
+                    #dm_command -= 60
```
ivalaginja: If you don't need this line, please delete

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,53 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
+    """
+    Creates a new flat map from the existing flat map for selected DM by adding a specified Zernike term with the
+    specified coefficient magnitude.
+    :param output_path: Path to output the new flatmap fits file. Default is hardware/boston/
+    :param file_name: Filename for new flatmap fits file. Default is
+            flat_map_volts_dm_<1 or 2>_zernike_<zernike index>_<coefficient value>V.fits
+    ;param dm_num: Which DM is this for?  Defaults to 1.
+    ;param zernike_index: What Zernike term are we applying, defaults to 4, which I think is focus, but should update if not
+    ;param zernike_coeff_volts: Coefficient of the Zernike term that we're adding in volts.  Defaults to +10, which for
+    ;                           most Zernike terms should be a P2V of 20.
```
ivalaginja: You want colons here, not semi-colons

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,53 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
+    catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
+
+
+def create_constant_flat_map(output_path, file_name=None, dm_num=1):
+    """
+    Creates a uniform flat map and outputs a new flatmap fits file.
+    :param output_path: Path to output the new flatmap fits file. Default is hardware/boston/
+    :param file_name: Filename for new flatmap fits file. Default is
+            flat_map_volts_dm_<1 or 2>_constant.fits
+    ;param dm_num: Which DM is this for?  Defaults to 1.
```
ivalaginja: you want a colon here, not a semi-colon

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,53 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
     max_volts = CONFIG_INI.getint("boston_kilo952", "max_volts")
     dm_command_data *= max_volts
-    catkit.util.write_fits(dm_command_data, output_path)
+    catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
+
+
+def create_constant_flat_map(output_path, file_name=None, dm_num=1):
+    """
+    Creates a uniform flat map and outputs a new flatmap fits file.
+    :param output_path: Path to output the new flatmap fits file. Default is hardware/boston/
```
ivalaginja: where is this default defined in the code?

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,53 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
     # Convert the dm command units to volts.
     max_volts = CONFIG_INI.getint("boston_kilo952", "max_volts")
     dm_command_data *= max_volts
-    catkit.util.write_fits(dm_command_data, output_path)
+    catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
```
ivalaginja: both here and for the Zernike maps - is it intended to *always* write them out? If so, maybe they should not go into the hicat repo, because that would cause confusion with files that need or don't need to be committed and pushed?

catkit/hardware/boston/DmCommand.py
grbrady: I uncomment this line and the line above when I want to apply a constant voltage offset to the DM command.  I'd prefer to leave them in so that I can remember where I do this.  Would appreciate suggestions, long term, on a better way to create a DM command with an adjusted constant voltage offset.

catkit/hardware/boston/DmCommand.py
grbrady: Fixed

catkit/hardware/boston/DmCommand.py
grbrady: Fixed

catkit/hardware/boston/DmCommand.py
grbrady: It's not, and probably shouldn't be.  Removed the comment.

catkit/hardware/boston/DmCommand.py
grbrady: To my knowledge the only way we can store and exchange raw voltage mirror commands is through the file system.  You're right that they probably shouldn't go into the repository and as was observed this isn't currently the default behavior.  When I use it from HiCat, the resulting file ends up in the data directory.

catkit/hardware/boston/DmCommand.py
jamienoss: These can't be left in. We can do that sort of thing in, e.g., a hicat script but not here in catkit hardware code.

Either, subtract it from the flatmap or do the subtraction (accounting for max volts) once the command was returned.

If it's a must, we can always refactor the ``bias`` arg to pass in the bias voltage.

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,50 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
+    """
+    Creates a uniform flat map and outputs a new flatmap fits file.
+    :param output_path: Path to output the new flatmap fits file.
+    :param file_name: Filename for new flatmap fits file. Default is
+            flat_map_volts_dm_<1 or 2>_constant.fits
+    :param dm_num: Which DM is this for?  Defaults to 1.
+    :return: None
+    """
+    if file_name is None:
+        file_name = "flat_map_volts_dm_" + str(dm_num) + "_constant.fits"
```
jamienoss: [revision]  ``file_name = f"flat_map_volts_dm_{dm_num}_constant.fits"``

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,50 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
+
+def add_zernike_to_flat_map(output_path, file_name=None, dm_num=1, zernike_index=4, zernike_coeff_volts=10):
+    """
+    Creates a new flat map from the existing flat map for selected DM by adding a specified Zernike term with the
+    specified coefficient magnitude.
+    :param output_path: Path to output the new flatmap fits file.
+    :param file_name: Filename for new flatmap fits file. Default is
+            flat_map_volts_dm_<1 or 2>_zernike_<zernike index>_<coefficient value>V.fits
+    :param dm_num: Which DM is this for?  Defaults to 1.
+    :param zernike_index: What Zernike term are we applying, defaults to 4, which I think is focus, but should update if not
```
jamienoss: [revision] Please update the _"which I think is focus, but should update if not"_.

catkit/hardware/boston/DmCommand.py
jamienoss: They're not though, not here at least, they'll end up in ``output_path`` which is the caller's responsibility.

jamienoss: We should have these funcs return the actual commands and make them write to file conditionally on ``output_path`` but not in this PR (unless it would be more useful to you @grbrady). I opened [CATKIT-10](https://jira.stsci.edu/browse/CATKIT-10) to address this.

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,50 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
+
+    flat_map_volts += (hicat_zernike_module.create_zernike(zernike_index, zernike_coeff_volts) + zernike_coeff_volts/2)
+    mask = catkit.util.get_dm_mask().astype('bool')
+    flat_map_volts *= mask
+    flat_map_volts_masked = flat_map_volts[mask]
+    flat_map_volts_masked[flat_map_volts_masked < 0] = 0
+    flat_map_volts[mask] = flat_map_volts_masked
+
+    if file_name is None:
+        file_name = f"flat_map_volts_dm_{dm_num}_zernike_{zernike_index}_{zernike_coeff_volts:.2f}V.fits"
```
jamienoss: We should probably be adding the mode and voltage to the resultant fits header rather than just relying on filename mangling.

If we would like this now I'm happy to add this if no one else cares to, just let me know. Otherwise, it can just be it's own PR ([CATKIT-11](https://jira.stsci.edu/browse/CATKIT-11)).

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,50 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
+    :return: None
+    """
+    flat_map_volts = get_flat_map_volts(dm_num)
+
+    flat_map_volts += (hicat_zernike_module.create_zernike(zernike_index, zernike_coeff_volts) + zernike_coeff_volts/2)
+    mask = catkit.util.get_dm_mask().astype('bool')
+    flat_map_volts *= mask
+    flat_map_volts_masked = flat_map_volts[mask]
+    flat_map_volts_masked[flat_map_volts_masked < 0] = 0
+    flat_map_volts[mask] = flat_map_volts_masked
```
jamienoss: [revision] This looks odd, I don't get the use of the ``flat_map_volts_masked``? Since you've already used the mask, the masked values will yield zeros and therefore never be < 0 so it doesn't matter that we clip the entire array.
```python
# Is this what you want?
flat_map_volts *= mask  # Zero masked values
flat_map_volts[flat_map_volts < 0] = 0  # Clip neg. values at 0.
```

catkit/hardware/boston/DmCommand.py
```diff
@@ -313,4 +316,50 @@ def create_flatmap_from_dm_command(dm_command_path, output_path, file_name=None,
...
+            flat_map_volts_dm_<1 or 2>_zernike_<zernike index>_<coefficient value>V.fits
+    :param dm_num: Which DM is this for?  Defaults to 1.
+    :param zernike_index: What Zernike term are we applying, defaults to 4, which I think is focus, but should update if not
+    ;param zernike_coeff_volts: Coefficient of the Zernike term that we're adding in volts.  Defaults to +10, which for
+    :                           most Zernike terms should be a P2V of 20.
+    :return: None
+    """
+    flat_map_volts = get_flat_map_volts(dm_num)
+
+    flat_map_volts += (hicat_zernike_module.create_zernike(zernike_index, zernike_coeff_volts) + zernike_coeff_volts/2)
```
jamienoss: [revision?] Why the ``/2``? Please add a brief comment, even if it's obvious but I'm just unfamiliar with the physics.

catkit/hardware/boston/DmCommand.py
```diff
@@ -4,9 +4,11 @@
 
 from astropy.io import fits
 import numpy as np
+from poppy import zernike
```
jamienoss: [revision] Remove as it's unused.

catkit/hardware/boston/DmCommand.py
```diff
@@ -4,9 +4,11 @@
...
 
 from catkit.config import CONFIG_INI
 import catkit.util
+from hicat.experiments.modules import zernike as hicat_zernike_module
```
jamienoss: [revision] I'm guessing you aliased this when you were also using ``poppy.zernike``. We should avoid aliasing and instead just use the full namespace, ``import hicat.experiments.modules.zernike`` and ``import poppy.zernike`` (but we no longer need that anyhow).

catkit/hardware/boston/DmCommand.py
grbrady: I added optional arguments up through the hierarchy to pass a constant voltage offset down to to_dm_command so that  the voltage can be biased up and down without resorting hard-coding and commenting out lines.

catkit/hardware/boston/DmCommand.py
grbrady: I went ahead and implemented this.  The files are not output unless output_path is specified and the functions now return the DM Command in volts, which is what I need when creating new flat maps.

catkit/hardware/boston/DmCommand.py
grbrady: Changed

catkit/hardware/boston/DmCommand.py
grbrady: Changed

catkit/hardware/boston/DmCommand.py
grbrady: I expect this is going to be extremely rarely used so it's probably fine to use filenames.  But feel free to add if you want.

catkit/hardware/boston/DmCommand.py
grbrady: I think I fell into a familiar Matlab pattern I use often for this, since Matlab doesn't have a =* operator.  Changed to what you suggest.

catkit/hardware/boston/DmCommand.py
grbrady: Zernike polynomials have a range from -1 to 1 (at least in the normalization used here I believe.)  Here I want a Zernike polynomial that has the same sign as the coefficient everywhere, so add in a factor of the coefficient to push everything to that side of zero.  I believe the /2 was erroneous.  It's possible that I empirically found that /2 gave the correct result, which could be the case if another normalization (like Noll's) is used.

Actually, on further consideration I just rewrote the code so that it pushes the Zernike to the correct side of zero based on the extreme value in the array instead of trying to figure it out based on the coefficient itself.  This should work regardless of normalization.

catkit/hardware/boston/DmCommand.py
grbrady: Ok, changed and updated the usage of it in the code.

catkit/hardware/boston/DmCommand.py
```diff
@@ -110,11 +110,11 @@ def to_dm_command(self):
                 if self.dm_num == 1:
                     flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm1")
                     flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
-                    dm_command += flat_map_volts[0].data
+                    dm_command += flat_map_volts[0].data + voltage_offset
                 else:
                     flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm2")
                     flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
-                    dm_command += flat_map_volts[0].data
+                    dm_command += flat_map_volts[0].data + voltage_offset
```
jamienoss: [revision] Hey, so this is not what we want, sorry. You've duplicated semantics. In my [original comment](https://github.com/spacetelescope/instrument-interface-library/pull/84#discussion_r422170559) I said that we could refactor the existing ``bias`` functionality not simply add another.

What you want can be achieved by just splitting the ``elif``:
```python
...
# Apply bias.
if self.bias:
  dm_command += self.bias_volts_dm1 if self.dm_num == 1 else self.bias_volts_dm2

if self.flat_map:
  ... # No changes to code but indentation.
```

There's a condition in the ``__init__()`` preventing the use of both flat maps and biasing that would need to be nuked also. Please check that there aren't other preventions.

jamienoss: @grbrady I'll do this for you.

catkit/hardware/boston/DmCommand.py
```diff
@@ -289,28 +305,92 @@ def convert_dm_image_to_command(dm_image, path_to_save=None):
...
+    bias_volts = CONFIG_INI.getint('boston_kilo952', f'bias_volts_dm{dm_num}')
+    mask = catkit.util.get_dm_mask()
+    dm_command_data = mask * bias_volts
+
+    if output_path is not None:
+        if file_name is None:
+            file_name = f"flat_map_volts_dm_{dm_num}_constant.fits"
+        catkit.util.write_fits(dm_command_data, os.path.join(output_path, file_name))
+
+    return dm_command_data
```
jamienoss, line 341: @grbrady I'm looking at this again and don't understand why this func exists at all.

Isn't this _just_ ``DmCommand(np.zeros(num_actuators), bias=True, flat_map=False).export_fits(...)``?

grbrady: Yes, I suppose that's true, however with my very, very basic understanding of DmCommand that line was pretty opaque to me until I dug into it a bit.  I get that the function I wrote above replicates functionality that is present in this DmCommand call redundantly.  However, for understanding of what's going on, I think there's value in wrapping the DmCommand call in a "create_constant_flap_map" or similarly named function.  Does that sound like a reasonable compromise?

ivalaginja: I see your point @grbrady, but I think in this case it might make more sense to just use the line Jamie suggested since the above function doesn't and anything but the clarification; you can always add a comment to clarify what it's doing, it can even be literally the docstring from the above function.

ivalaginja:
@mperrin @jamienoss can we close this one too then or do we need something of it?

mperrin:
Looks to me like this PR is not much about phase retrieval itself, but instead about using phase retrieval results to adjust the DM flat maps? If so I am not the person to ask if any of it is still useful. You could ask Matt or Remi perhaps. But my sense is that the subsequent DM calibration efforts went in a different direction than this code.

jamienoss:
@RemiSoummer, @mmaclay See @mperrin [comment above](https://github.com/spacetelescope/catkit/pull/84#issuecomment-784207543).

PR #107, opened 2020-07-21
HICAT-864 Fix compound quantities for all

jamienoss:
FYI I'm not yet sure if this actually fixes HICAT-864.

@grbrady has also experienced this with the phase retrieval code. This PR will fix that issue.

### DIFF
--- a/catkit/catkit_types.py
+++ b/catkit/catkit_types.py
@@ -52,7 +52,29 @@ SinSpecification = namedtuple("SinSpecification", "angle, ncycles, peak_to_valle
 
 # Create shortcuts for using Pint globally.
 units = UnitRegistry()
-quantity = units.Quantity
+
+
+class Quantity(units.Quantity):
+    """ Wrapper for pint.Quantity to avoid compounding quantities.
+
+        https://pint.readthedocs.io/en/stable/developers_reference.html#pint.Quantity
+    """
+    # Add this alias due to __news__() shadowing of the above global units.
+    _units = units
+
+    def __new__(cls, value, units=None):
+        if isinstance(value, cls._units.Quantity):
+            if isinstance(units, str):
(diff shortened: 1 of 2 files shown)


catkit/catkit_types.py
```diff
@@ -52,7 +52,29 @@ class ImageCentering(Enum):
...
+class Quantity(units.Quantity):
+    """ Wrapper for pint.Quantity to avoid compounding quantities.
+
+        https://pint.readthedocs.io/en/stable/developers_reference.html#pint.Quantity
+    """
+    # Add this alias due to __news__() shadowing of the above global units.
+    _units = units
+
+    def __new__(cls, value, units=None):
+        if isinstance(value, cls._units.Quantity):
```
jamienoss, line 66: The base issue is that ``pint`` quantities are dynamically defined belonging to an instance of ``pint.UnitRegistry()``. This results in ``if isinstance(value, cls._units.Quantity)`` _always_ returning ``False`` on Windows due to the multiprocess spawn.

Updating ``_units = units`` with that on the spawn works only so far and then runs into the issue that ``pint`` quantities from diff registries aren't compatible with one another. We could just nuke an old instance and create a new but things are starting to become a slippery rabbit hole.

Another solution would be to not use ``pint`` from the parent process which includes it's use from scripts and class instantiations (From scripts), class/func defaults and other "compile"/import time usages. This would keep all ``quantity`` instances to a single registry and also from that of the spawned child process.

Without the latter above-mentioned condition, it actually becomes impossible to even create a literally defined subclass of a pint quantity, it has to dynamically created, which is even more unnecessary-ness.

jamienoss:
In light of https://github.com/spacetelescope/catkit/pull/107#discussion_r509613520 and in favor of https://github.com/spacetelescope/catkit/pull/133 I'm closing this PR.

PR #108, opened 2020-07-22
HICAT-657 Refactor Zwo interface and add emulation

jamienoss:
Moved from #46 

- [x] [CATKIT-21](https://jira.stsci.edu/browse/CATKIT-21) Catch zwo RuntimeError: Exposure status: 3 and raise CLOSE CAM IN SHARPCAP!!!
- [x] Tested on hardware in conjunction with https://github.com/spacetelescope/hicat-package/pull/319.

We'll want to wait till https://github.com/spacetelescope/hicat-package/pull/162 is merged before merging this so as to save others from having to smash them together - I can do it.

https://github.com/spacetelescope/hicat-package/pull/319 downstream PR.


catkit/emulators/ZwoCamera.py
```diff
@@ -0,0 +1,157 @@
+import numpy as np
+
+import zwoasi
+
+from hicat.config import CONFIG_INI
+from hicat.hardware import testbed_state
+import hicat.simulators
```
mperrin: I was under the impression we desired to avoid hicat-specific code mashed into catkit?

catkit/emulators/ZwoCamera.py
```diff
@@ -0,0 +1,157 @@
...
+
+from catkit.interfaces.Instrument import SimInstrument
+import catkit.hardware.zwo.ZwoCamera
+
+
+"""SIMULATED Implementation of Hicat.Camera ABC that provides interface and context manager for using ZWO cameras."""
+
```
mperrin: Can delete this comment string

catkit/emulators/ZwoCamera.py
```diff
@@ -0,0 +1,157 @@
...
+        # Adjust count rate to compensate for poppy normalizing to 1.0 in entrance pupil by default.
+        exit_pupil_flux_correction_factor =  CONFIG_INI.getfloat('data_simulator', 'simulator_direct_exit_vs_entrance_pupil_flux')
+        counts_per_microsec /= exit_pupil_flux_correction_factor
+
+        # Apply flux normalization and exposure time scaling to the output image
+        exposure_time = self.control_values[self.ASI_EXPOSURE]
+        image = image_hdulist[0].data * counts_per_microsec * exposure_time
+
+        return image
```
jamienoss: Commits prior to 75427d97e906d6146f3c644976464a1eed96b4b8 returned ``image.astype(np.dtype(np.int<X>))``. 

This actually causes the following test to fail: https://github.com/spacetelescope/hicat-package/blob/develop/hicat/simulators/tests/test_sim_hardware.py#L135-L146
```python
def test_nd_filter_reduces_flux(ensure_testbed_sim_mode):
    camera_name = CONFIG_INI.get("testbed", "imaging_camera")

    with hicat.simulators.ZwoCamera(camera_name) as camera:
        # Test in 'data mode', in which images are handed back without any files written to disk
        images_1ms, metadata = camera.take_exposures(1000, 1, return_metadata=True)

        # Check that adding an ND filter reduces the apparent flux
        hicat.simulators.optics_simulator.nd_filter = '9_percent'
        (im_1ms_nd9p,) = camera.take_exposures(1000, 1)
    assert np.isclose(images_1ms[0].sum() * 0.09, im_1ms_nd9p.sum(),
                      rtol=0.01), "Image total counts didn't scale as expected with 9% ND filter"
```

jamienoss: Without casting/rounding to ``int`` ``images_1ms[0].sum() = ~2337231.9`` (as the old sim yields). When casting ``images_1ms[0].sum() = 2162269``.

With 9% filter the results are:
210350.8 (old/current) and 153302 (this PR).

Implying that the sim yields the following average fp "noise" levels.
(2337231.919896499 - 2162269)/712/712 = 0.34513263772033753
(210350.87279068487-153302)/712/712 = 0.11253486142588702

jamienoss: @mperrin I have no idea what assumptions go into calibrating/matching the simulator to the testbed and was wondering if this could imply some sort of mismatch or perhaps just some other implication?

I was thinking along the lines that we calibrate the sim from camera data which returns ints and whether us not casting to ints implies that the physics being simulated "behind the cameras" by Poppy isn't "seeing" the correct values? Though I also get that we average multiple exposures on the testbed to regain this sort of thing but don't for the sim, which could explain away the entire thing.

Anyhoo, possible food for thought.

catkit/emulators/ZwoCamera.py
```diff
@@ -0,0 +1,157 @@
...
+    def stop_video_capture(self):
+        pass
+
+    def stop_exposure(self):
+        pass
+
+    def set_image_type(self, image_type):
+        self.image_type = image_type
+
+    def capture(self, initial_sleep=0.01, poll=0.01, buffer=None, filename=None):
```
mperrin, line 79: The entire `capture` function is full of HICAT specifics. I don't think any of this belongs in catkit.

catkit/emulators/ZwoCamera.py
mperrin: Hi Jamie, I'm not sure I follow the concerns here. The task in matching the simulator optical model to the testbed is almost entirely about PSF _shape_ (speckle pattern, spatial scale, relationship of DM influence functions to affected spatial regions in the image plane), and not about intensity.  

In any case the simulation of the physics of the camera, including the readout and casting to ints and floats, is all rigorously implemented in an improved fashion in https://github.com/spacetelescope/hicat-package/pull/162

catkit/emulators/ZwoCamera.py
```diff
@@ -0,0 +1,157 @@
...
+        # set_roi_format --> _set_roi_format : 
+        # check for all the errors 
+        # runs zwolib.ASISetROIFormat(id_, width, height, bins, image_type)
+        # set_roi_start_position --> _set_start_position :
+        # runs zwolib.ASISetStartPos(id_, start_x, start_y)
+        pass  # according to Marshall
+
+
+class ZwoCamera(SimInstrument,  catkit.hardware.zwo.ZwoCamera.ZwoCamera):
+    """ Now we use poppy to take images."""
```
mperrin: This docstring is unclear and not well descriptive...

catkit/emulators/ZwoCamera.py
mperrin: And relatedly the test that you refer to as failing here is also revised in https://github.com/spacetelescope/hicat-package/pull/162 too.

catkit/emulators/ZwoCamera.py
jamienoss: You're correct it doesn't, however, this is at least a clean step in one direction that we can back out later.

However, how can we do this otherwise? Perhaps just passing in a ref to the sim object upon instantiation? E.g., like we do with the DM emulation. 🤔 I think that might do it.
BTW The reason I made ``get_camera_mappings()`` a classmethod (which we can trivially make static) is so that we can call it in ``hicat.testbed`` so as to push the hicat stuff back into hicat.

catkit/emulators/ZwoCamera.py
jamienoss: 👍  If you're cool with it I am. I just wanted to bring it to your attention in case it meant anything to you.

catkit/emulators/ZwoCamera.py
mperrin: Yes, I was thinking either similar to how we do the DM emulation is one option; the other could be having this on method as an abstract method which we have to subclass and override with a specific implementation in catkit. 

If we wanted to do it similar to the DM implementation, one option would be to pass in to `__init__` a function handle `get_noiseless_count_rate_image()` or something along those lines. That function would include what's currently in lines 88-116 of this code (or equivalent thereof from the PR 162), and would return an image in count rate units, i.e. equivalent of `image_hdulist[0].data * counts_per_microsec` in the current code. Then this function would apply the readout time and the camera noise model. That would make it like this:

```python
def capture(self, initial_sleep=0.01, poll=0.01, buffer=None, filename=None):
    count_rate_image = self.get_noiseless_count_rate_image()

    image = count_rate_image * exposure_time
    # add noise model here to image
    return image
```

I think that's probably the cleanest way to do this, and keeps it pretty comparable to how we do the DM model. What do you think?

catkit/emulators/ZwoCamera.py
mperrin: Yeah much appreciated. Thanks for asking - always good to point out curiosities like this.   Camera quantization does make a noticeable though small impact on the images, but is not a limiting factor in the overall SNR, as long as we have good exposure times and sufficient number of exposures. The photon noise and read noise terms are larger contributors to the overall noise budget for the most part.

catkit/emulators/tests/test_web_power_switch.py
```diff
@@ -25,13 +25,13 @@ def test_all_off():
 @pytest.mark.usefixtures("dummy_config_ini", "dummy_testbed_state")
 def test_individual_on():
     switch = WebPowerSwitch(config_id="web_power_switch")
-    switch.turn_on("dm1_outlet")
+    switch.turn_on("fpm_led")
```
jamienoss: Updated (I think - if I can remember correctly) because I updated the fake config.ini copy.

jamienoss:
@mperrin I haven't taken a final look over this myself yet, however, I wanted to throw it back on your radar. 

I'll take a look myself on Monday but it should be good to merge next week.

jamienoss:
Upstream HICAT testing failing as expected.

PR #114, opened 2020-08-25
Testing bufix for unit error.

julesfowler:
As noted by Marshall and Iva, some of our calibration scripts fail because we pick up nested exposure times. 
This introduces a function to strip away all but the bottom layer of units, so that we don't keep trying to pass these nested quantities in.

This problem is also being addressed in https://github.com/spacetelescope/instrument-interface-library/pull/107, but since that has stalled and this continues to be an issue I figured I would throw up this less elegant solution in the meantime. 

This has been tested on hardware, and runs through for `take_sampling_data.py`.

### DIFF
--- a/catkit/hardware/zwo/ZwoCamera.py
+++ b/catkit/hardware/zwo/ZwoCamera.py
@@ -1,7 +1,6 @@
 from astropy.io import fits
 import numpy as np
 import logging
-import os
 import zwoasi
 import sys
 
@@ -218,6 +217,9 @@ class ZwoCamera(Camera):
         """Applies control values found in the config.ini unless overrides are passed in, and does error checking."""
 
         # Load values from config.ini into variables, and override with keyword args when applicable.
+        # Convert exposure time to contain units if not already a Pint quantity.
+        if type(exposure_time) is int or type(exposure_time) is float:
+            exposure_time = quantity(exposure_time, units.microsecond)
         subarray_x = subarray_x if subarray_x is not None else CONFIG_INI.getint(self.config_id, 'subarray_x')
         subarray_y = subarray_y if subarray_y is not None else CONFIG_INI.getint(self.config_id, 'subarray_y')
         width = width if width is not None else CONFIG_INI.getint(self.config_id, 'width')
--- a/catkit/hardware/zwo/ZwoCamera.py
+++ b/catkit/hardware/zwo/ZwoCamera.py
@@ -175,7 +175,8 @@ class ZwoCamera(Camera):
                                     height=height, gain=gain, full_image=full_image, bins=bins)
 
         # Create metadata from testbed_state and add extra_metadata input.
-        meta_data = [MetaDataEntry("Exposure Time", "EXP_TIME", exposure_time.to(units.microseconds).m, "microseconds")]
+        exposure_time = self._strip_units(exposure_time)
+        meta_data = [MetaDataEntry("Exposure Time", "EXP_TIME", exposure_time.to(units.microseconds).magnitude, "microseconds")]
         meta_data.extend(testbed_state.create_metadata())
         meta_data.append(MetaDataEntry("Camera", "CAMERA", self.config_id, "Camera model, correlates to entry in ini"))
         meta_data.append(MetaDataEntry("Gain", "GAIN", self.gain, "Gain for camera"))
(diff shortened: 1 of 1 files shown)


catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -1,7 +1,6 @@
 from astropy.io import fits
 import numpy as np
 import logging
-import os
```
julesfowler, line 4: This import wasn't being used, so I figured I would zap it.

catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -196,6 +196,39 @@ def just_take_exposures(self, exposure_time, num_exposures,
 
         return img_list, meta_data
 
+    def _strip_units(self, nested_quantity):
+        """ Sometimes we get compound quantities that look like :
+        >>> 100 <Unit('microsecond')> <Unit('microsecond')>
```
mperrin, line 200: So, we're getting malformed inputs, and the approach here seems to be to adapt catkit so that it can handle these? Dropping off the doubled units?  That may work as a workaround, but it's not actually addressing the root cause. It doesn't make sense to get an exposure time with units of time squared, so the calling code in HiCAT should not be producing that. 

In other words, is this perhaps a bug that is better fixed by correcting the root cause in hicat, rather than adapting catkit code to handle bad inputs?

catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -196,6 +196,39 @@ def just_take_exposures(self, exposure_time, num_exposures,
...
+        This function will strip off the extra layers of units and return :
+        >>> 109 <Unit('microsecond')>
```
ivalaginja: a typo here

catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -196,6 +196,39 @@ def just_take_exposures(self, exposure_time, num_exposures,
 
         return img_list, meta_data
 
+    def _strip_units(self, nested_quantity):
```
mperrin: [revision] Let's rename this for clarity to `_strip_redundant_units` or `_fix_malformed_units`. Some name that will indicate that it does not strip off all units information, just incorrectly specified information.

mperrin: I would also add a note in the text that this is a bug workaround rather than a real fix, and link to some of the relevant GitHub or JIRA issues or at least reference their numbers. That will help make it clear later when/if someone is to address this more carefully.

catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -196,6 +196,39 @@ def just_take_exposures(self, exposure_time, num_exposures,
...
+        >>> 109 <Unit('microsecond')> 
+
+        Parameters
+        ----------
+        nested_quantity : catkit_types.quantity
+            A potentially nested quantity.
+        
+        Returns
+        -------
+        stripped_quantity : catkit_types.quantity
```
mperrin: Perhaps rename parameters to `quantity_to_check` and `valid_quantity`?

catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -196,6 +196,39 @@ def just_take_exposures(self, exposure_time, num_exposures,
...
+            A quantity with only one unit and scalar magnitude.
+        """
+        try:
+            n = 0
+            has_units = 'magnitude' in dir(nested_quantity.magnitude)
+            stripped_quantity = nested_quantity
+            while has_units and n < 10:
+                stripped_quantity = stripped_quantity.magnitude
+                has_units = 'magnitude' in dir(stripped_quantity.magnitude)
+                n += 1
```
mperrin: [revise] This part should use `log.warning` to emit a log message about what it's doing. Along the lines of
`log.warning(f'Malformed redundant units in input {nested_quantity}. Correcting to {stripped_quantity}')`

This is both to be explicit that a parameter input to this function is being changed, and (importantly) to put a flag in the log file that we can use to track down the cases in which this is happening, determine the root cause, and fix it.

catkit/hardware/zwo/ZwoCamera.py
ivalaginja: the note on this being a workaround/link to Jira issue is still missing, otherwise this looks good

jamienoss:
This PR only addresses a symptom and not all scenarios where compound quantities aren't desired, e.g., other sections of code that write these values to file etc. A more complete solution would have been to subclass pint quantity and override ``magnitude()`` adding the strip functionality in there. That would work if it weren't for issues with subclassing pint quantities as indicated [here](https://github.com/spacetelescope/catkit/pull/107#discussion_r509613520).

For the above reason and in favor of https://github.com/spacetelescope/catkit/pull/133 I'm closing this PR.

PR #119, opened 2020-09-14
Update setup.py

ivalaginja:
This has been confusing the crap out of me while working with different conda environments.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='catkit',
-      version='0.5.2',
+      version='0.13.0',
       description='Python API wrapping hardware and device drivers for the Optics Lab.',
       url='https://github.com/spacetelescope/instrument-interface-library',
       author='Makidon Optics Lab',


jamienoss:
I felt like changing this too, a while back. Just be aware that this doesn't work with a dev install, which is what we tend to use. ``conda list`` will show you only that when installed not when a new branch/tag is checked out.

ivalaginja:
@jamienoss are you sure? I just tried it on spacetelescope/pastis (branch beyond/package-setup), where I also install with dev and there conda list shows me what's in the setup.py

jamienoss:
Lol not now I'm not. I had tried it again too, with this very branch and it didn't work as it hadn't done before.

What exactly did you do?

ivalaginja:
I did `cd instrument-interface-library` and then ran this:

Screenshot 2020-09-14 at 16 37 42

The `conda list` shows me `catkit 0.13.0` now.

ivalaginja:
Of course this would not work if v0.13.0 or anything before is checked out since the version of the `setup.py` it contains won't have this update included. Which also means that this needs to be updated with every new release otherwise it's not fixed at all, and this PR here won't fix it either.

jamienoss:
@ivalaginja ok, that's what I thought you would have done. So if you re-install it as you did then it will obviously work. However, the point of the dev install is that you don't reinstall it every time you check out a new branch.

ivalaginja:
Correct, but it will at least show up as the most recent release from which you installed initially as opposed to one from over a year ago, as long as this gets updated along with every release. So as I said, this PR won't fix anything per se, but might reduce confusion as to why your catkit installation shows up as older than this frickin' pandemic.

jamienoss (approved):

jamienoss:
@ivalaginja For sure, I wasn't saying that we don't do this just to be mindful of its limitations, that this too can easily yield the same confusion as the scenario that lead you to open this PR.

PR #131, opened 2020-10-20
Replace laser_config_id parameter with actual wavelength

ivalaginja:
Currently, the various classes in `segmented_dm_command` require a `laser_config_id` parameter in order to read the wavelength of the system from the testbed configfile. This wavelength information is only used in the poppy plotting functions of the segmented dm command and do not influence the command being sent to the DM (simulated or hardware) directly.

This very small PR makes the input parameter be the wavelength directly, it seemed to me it would make more sense this way. @kjbrooks this would influence https://github.com/spacetelescope/hicat-package/pull/340 and also https://github.com/spacetelescope/jost-package/pull/15.

### DIFF
--- a/catkit/hardware/iris_ao/segmented_dm_command.py
+++ b/catkit/hardware/iris_ao/segmented_dm_command.py
@@ -4,10 +4,10 @@ segmented DM hardware as a command.
 
 
 This module can be used to create a command in the following way:
-    poppy_obj = PoppySegmentedCommand(global_coefficients, dm_config_id, laser_config_id)
+    poppy_obj = PoppySegmentedCommand(global_coefficients, dm_config_id, wavelength)
     command = poppy_obj.to_dm_list()
     iris_command = segmented_dm_command.load_command(command, dm_config_id,
-                                                     laser_config_id,
+                                                     wavelength,
                                                      testbed_config_id,
                                                      apply_flat_map=True)
     iris_command.display()
@@ -31,7 +31,7 @@ from catkit.config import CONFIG_INI
 from catkit.hardware.iris_ao import util
 
 
-class SegmentedAperture():
+class SegmentedAperture:
     """
     Create a segmented aperture with Poppy using the parameters for the testbed
     and segmented DM from the config.ini file.
@@ -42,9 +42,9 @@ class SegmentedAperture():
 
     :param dm_config_id: str, name of the section in the config_ini file where information
                          regarding the segmented DM can be found.
-    :param laser_config_id: str, name of the section in the config_ini file where information
-                         regarding the laser can be found.
-    :attribute wavelength: int, wavelength of the laser being used
+    :param wavelength: float, wavelength in nm of the poppy optical system used for
+                        (extremely oversimplified) focal plane simulations
+    :attribute wavelength: float, wavelength in nm of the poppy optical system
     :attribute outer_ring_corners: bool, whether or not the segmented aperture includes
                                    the corner segments on the outer-most ring. If True,
                                    corner segments are included. If False, they are not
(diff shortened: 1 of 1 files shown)


ivalaginja:
@jamienoss just FYI as we haven't started using these command objects on HiCAT yet, but hopefully we will soon. Some  more cleanup will follow as we move along.

ivalaginja:
@jamienoss *help*! I thought we had not added tests that include the segmented dm command yet, but it seems like I have broken tests that deal with `testbed`. This is not a super urgent PR, but would you mind looking at it when you get a moment?

jamienoss:
@ivalaginja This is all cool. These are the upstream hicat tests that are broken until https://github.com/spacetelescope/hicat-package/pull/319 gets merged.

Sorry, I should have mentioned this, I've gotten used to being the only dev of catkit.

ivalaginja:
Ah ok cool @jamienoss  thanks! This means this can be reviewed as-is @kjbrooks .

jamienoss:
hicat-package/develop is on catkit v0.15.1 so the tests should all pass now.

catkit/hardware/iris_ao/segmented_dm_command.py
```diff
@@ -225,9 +224,9 @@ class SegmentedDmCommand(SegmentedAperture):
                          that you are defining
     """
 
-    def __init__(self, dm_config_id, laser_config_id, testbed_config_id, apply_flat_map=False):
+    def __init__(self, dm_config_id, wavelength, testbed_config_id, apply_flat_map=False):
```
kjbrooks, line 227: This is a minor thing but I recommend reordering this just for readability:
``` def __init__(self, dm_config_id, testbed_config_id, wavelength, apply_flat_map=False)```

File: catkit/hardware/iris_ao/segmented_dm_command.py [hunk 2]
```diff
@@ -4,10 +4,10 @@
 
 
 This module can be used to create a command in the following way:
-    poppy_obj = PoppySegmentedCommand(global_coefficients, dm_config_id, laser_config_id)
+    poppy_obj = PoppySegmentedCommand(global_coefficients, dm_config_id, wavelength)
```
kjbrooks, line 7: This would be in all places where `wavelength` is used, but I recommend using `laser_wavelength` as a variable name to make this a little more explicit.

catkit/hardware/iris_ao/segmented_dm_command.py
```diff
@@ -579,8 +578,8 @@ class PoppySegmentedCommand(SegmentedAperture):
                                 [piston, tip, tilt, defocus, ...] (Noll convention)
     :param dm_config_id: str, name of the section in the config_ini file where information
                          regarding the segmented DM can be found.
-    :param laser_config_id: str, name of the section in the config_ini file where information
-                            regarding the laser can be found.
+    :param wavelength: wavelength: float, wavelength in nm of the poppy optical system used for
+                        (extremely oversimplified) focal plane simulations
```
kjbrooks, line 582: I am not sure it is necessary to say "extremely oversimplified", maybe just "simplified"?

kjbrooks (approved):
I just had a few small comments but nothing that is absolutely necessary to change.

catkit/hardware/iris_ao/segmented_dm_command.py
ivalaginja: "Simplified" to me is when we put in a thin lens in a simulation to approximate an out-of focus optic, "extremely simplified" is to me when we literally ignore the *entire* optical setup and just do one single Fourier transform from the pupil to a focal plane, like in this case with the PSF display ^^

kjbrooks: Okay, but those words don't necessarily mean exactly the same thing to others. Maybe point to the poppy documentation so that people can understand that for themselves?

PR #133, opened 2020-10-21
HICAT-466 use astropy.units

jamienoss:
Resolves HICAT-466
Resolves HICAT-864

- [x] Requires https://github.com/spacetelescope/hicat-package/pull/393 to not break downstream hicat tests.
- [x] Tested on hardware (2020-10-22T11-47-53)

### DIFF
--- a/catkit/catkit_types.py
+++ b/catkit/catkit_types.py
@@ -73,6 +73,25 @@ class Quantity(astropy.units.Quantity):
     def __round__(self, n=None):
         return self.round(decimals=n)
 
+    # See https://docs.astropy.org/en/stable/units/quantity.html#subclassing-quantity
+    def __quantity_subclass__(self, unit):
+        """
+        Overridden by subclasses to change what kind of view is
+        created based on the output unit of an operation.
+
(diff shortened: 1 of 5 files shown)


environment.yml
```diff
@@ -12,7 +12,6 @@ dependencies:
   - astropy>=1.3
   - numpy>=1.15.0
   - pillow>=4.0.0
-  - conda-forge::pint<=0.9.0
```
jamienoss, line 15: This will need to be removed from hicat also.

catkit/catkit_types.py
```diff
@@ -50,9 +51,50 @@ class ImageCentering(Enum):
...
+        unit : UnitBase
+            The unit for which the appropriate class should be returned
+
+        Returns
+        -------
+        tuple :
+            - `Quantity` subclass
+            - bool: True if subclasses of the given class are ok
+        """
+        return Quantity, True
```
jamienoss, line 93: This is a verbatim copy and works because both classes are named ``Quantity``.

PR #135, opened 2020-10-23
HICAT-951: IrisAO emulator

ivalaginja:
[JIRA: HICAT-951](https://jira.stsci.edu/browse/HICAT-951)

Create an IrisAO emulator.

Needed for https://github.com/spacetelescope/hicat-package/pull/395


ivalaginja:
@jamienoss the problem I ran into is that `subprocess.CREATE_NEW_PROCESS_GROUP` in line https://github.com/spacetelescope/catkit/blob/a31c32de402989ffcd4324c187b9752741814560/catkit/hardware/iris_ao/IrisAoDmController.py#L102
 apparently only exists in Windows? When I set `creationflags=None` I manage to run the `iris_ao.apply_flat_map.py` (without checking for correct results, but it runs).

jamienoss:
Moved to #136

PR #139, opened 2020-11-01
Spare IrisAO segment numbers testing

ivalaginja:
Upstream for https://github.com/spacetelescope/hicat-package/pull/340

These are changes that will provide us with correctly controllable hardware, which we can then use as a baseline to do any refactoring in the segmented DM controls.

PR #140, opened 2020-11-03
HiCAT-951: IrisAO segment ID

ivalaginja:
Some follow up to #139, and goes with HiCAT PR https://github.com/spacetelescope/hicat-package/pull/406

### DIFF
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -61,7 +61,7 @@ class PoppyIrisAODM(poppy.dms.HexSegmentedDeformableMirror):
         return new_data
 
     def relax(self):
-        zeros_list = catkit.hardware.iris_ao.util.create_zero_list(37)
+        zeros_list = catkit.hardware.iris_ao.util.create_zero_list(self.number_of_segments)
         zero_dict = catkit.hardware.iris_ao.util.create_dict_from_list(zeros_list)
         self.set_surface(zero_dict)
 
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -4,7 +4,7 @@ import astropy.units as u
 import poppy
 
 from catkit.hardware import testbed_state
-import catkit.hardware.iris_ao.iris_ao_controller
+from catkit.hardware.iris_ao.iris_ao_controller import IrisAoDmController
 import catkit.hardware.iris_ao.segmented_dm_command as segmented_dm_command
 import catkit.hardware.iris_ao.util
 from catkit.interfaces.Instrument import SimInstrument
@@ -118,5 +118,5 @@ class PoppyIrisAOEmulator:
         pass
 
 
-class PoppyIrisAoDmController(SimInstrument, catkit.hardware.iris_ao.iris_ao_controller.IrisAoDmController):
+class PoppyIrisAoDmController(SimInstrument, IrisAoDmController):
     instrument_lib = PoppyIrisAOEmulator
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -1,6 +1,7 @@
 import os
 
 import astropy.units as u
+import numpy as np
 import poppy
 
 from catkit.hardware import testbed_state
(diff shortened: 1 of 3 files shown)


catkit/emulators/iris_ao_controller.py
```diff
@@ -114,7 +114,7 @@ def Popen(self,
         return self
 
     def write(self, buffer):
-        if self.disable_hardware:
+        if not self.disable_hardware:
```
jamienoss: @ivalaginja Why are we changing this?

ivalaginja: Because it makes things work. I have no idea how and why, but in this form it actually takes the commands and communicates with the simulated hardware. There might be a better way to fix this?

jamienoss: If it's not working then it's because ``disable_hardware=True`` when you want it to be ``False``. This piece of logic was correct as is. If ``disable_hardware == True`` we don't want to write/talk to the hardware so this func just returns.

catkit/hardware/iris_ao/segmented_dm_command.py
```diff
@@ -10,7 +10,7 @@
                                                      wavelength,
                                                      testbed_config_id,
                                                      apply_flat_map=True,
-                                                     filename_flat=FLAT_COMMAND_FILE)
+                                                     filename_flat='repo-path/hardware/iris-ao/CustomFLAT.ini')
```
jamienoss, line 13: [revision] ``None`` or ``''``.

catkit/emulators/iris_ao_controller.py
jamienoss: Fixed in https://github.com/spacetelescope/catkit/pull/141

catkit/hardware/iris_ao/segmented_dm_command.py
ivalaginja: @kjbrooks suggested I just put an example in here, since it's a comment to help guide the user.

catkit/emulators/iris_ao_controller.py
```diff
@@ -50,7 +60,9 @@ def invert_data(data):
         return new_data
 
     def relax(self):
-        self.set_surface(self.relaxed_poppy_surface)
+        zeros_list = catkit.hardware.iris_ao.util.create_zero_list(37)
```
jamienoss: [revision] 37 -> ``self.number_of_segments``. @ivalaginja

catkit/emulators/iris_ao_controller.py
ivalaginja: So should I revert this or leave as is?

jamienoss: This needs to be reverted.

catkit/hardware/iris_ao/segmented_dm_command.py
jamienoss: Then do so in the docstring, not the actual default arg.

ivalaginja: it's not an arg though, it's a docstring

jamienoss: Hahahaha, I hadn't noticed that. Forget I said anything - these are not the droids you're looking for.

catkit/emulators/iris_ao_controller.py
```diff
@@ -36,8 +37,17 @@ def __init__(self, mcf_filename, custom_flat_filename, mirror_serial, **super_kw
...
+            # Addition of dicts taken from old JOST version here:
+            # https://github.com/spacetelescope/jost-package/blob/aa7cc5517bb474eed971db0f4007ce5fc1226fec/jost/DM_functions.py#L453
+            dm_surface = {seg: tuple(np.asarray(self.relaxed_poppy_surface.get(seg, (0., 0., 0.))) + np.asarray(dm_command_dict.get(seg, (0., 0., 0.)))) for
+                          seg in set(self.relaxed_poppy_surface) | set(dm_command_dict)}
+            return dm_surface
+
         # Setting the simulated IrisAO means setting each actuator individually
-        for seg, values in new_surface.items():
+        for seg, values in convert_command_to_poppy_surface(new_surface).items():
             self.set_actuator(seg-1, values[0] * u.um, values[1] * u.mrad, values[2] * u.mrad)  # TODO: double-check the -1 here, meant to correct for different segment names
```
jamienoss, line 51: This could be improved upon but doesn't need to hold up this PR.

jamienoss (approved):

jamienoss:
@ivalaginja Ignore the out of date notice and feel free to merge post tests.

catkit/emulators/iris_ao_controller.py
ivalaginja: agreed

ivalaginja: Also, there is actually a bug in this which I will fix in [CATKIT-50](https://jira.stsci.edu/browse/CATKIT-50).

PR #144, opened 2020-11-05
CATKIT-50: Fix inversion of tip and tilt command

ivalaginja:
[CATKIT-50](https://jira.stsci.edu/browse/CATKIT-50)

### DIFF
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -48,7 +48,13 @@ class PoppyIrisAODM(poppy.dms.HexSegmentedDeformableMirror):
 
         # Setting the simulated IrisAO means setting each actuator individually
         for seg, values in convert_command_to_poppy_surface(new_surface).items():
-            self.set_actuator(seg-1, values[0] * u.um, values[2] * u.mrad, values[1] * u.mrad)  # TODO: double-check the -1 here, meant to correct for different segment names
+            # The IrisAO hardware component and the poppy hex DM object have flipped
+            # x/y coordinate axes, so we need to feed what is called "tilt" on the IrisAO
+            # into the "tip" argument from poppy and vice versa.
+            piston = values[0] * u.um
+            tip = values[2] * u.mrad
+            tilt = values[1] * u.mrad
+            self.set_actuator(seg-1, piston, tip, tilt)  # TODO: double-check the -1 here, meant to correct for different segment names
 
     @staticmethod
     def invert_data(data):
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -48,7 +48,7 @@ class PoppyIrisAODM(poppy.dms.HexSegmentedDeformableMirror):
 
         # Setting the simulated IrisAO means setting each actuator individually
         for seg, values in convert_command_to_poppy_surface(new_surface).items():
-            self.set_actuator(seg-1, values[0] * u.um, values[1] * u.mrad, values[2] * u.mrad)  # TODO: double-check the -1 here, meant to correct for different segment names
+            self.set_actuator(seg-1, values[0] * u.um, values[2] * u.mrad, values[1] * u.mrad)  # TODO: double-check the -1 here, meant to correct for different segment names
 
     @staticmethod
     def invert_data(data):


mperrin:
Since this is subtle, and potentially confusing later for someone else looking at this code, let's please add a comment to explain what's going on and that this transposition is intentional.

catkit/emulators/iris_ao_controller.py
```diff
@@ -48,7 +48,7 @@ def convert_command_to_poppy_surface(dm_command_dict):
 
         # Setting the simulated IrisAO means setting each actuator individually
         for seg, values in convert_command_to_poppy_surface(new_surface).items():
-            self.set_actuator(seg-1, values[0] * u.um, values[1] * u.mrad, values[2] * u.mrad)  # TODO: double-check the -1 here, meant to correct for different segment names
+            self.set_actuator(seg-1, values[0] * u.um, values[2] * u.mrad, values[1] * u.mrad)  # TODO: double-check the -1 here, meant to correct for different segment names
```
jamienoss: Instead of _just_ a comment we can expand the code a little also, i.e.,
```python
piston_stroke = values[0] * um
tip = values[1] * umrad
tilt = values[2] * umrad
self.set_actuator(seg-1, piston_stroke, tip, tilt)
```
(correcting for which ever way around it actually is 🤷 )

ivalaginja: I like this

jamienoss: We then only need to comment about what expects what in which order, rather than also commenting on index semantics.

ivalaginja:
@mperrin would you have an idea how I could check the PTT sign in this? For tip and tilt I could measure with the 4D what a particular tip command does and then infer the correct (sub)PSF shift in the focal plane image in the simulator I think? Does that make sense? This would require me to trace it through the optical system though.

ivalaginja:
I could also open a ticket to check this by comparing the hardware to sim once the IrisAO is on the table, that might be easier.

mperrin:
Iva - for checking the sign, use your letter F pattern but with different and smaller motions per segment so it doesn’t phase wrap. For instance tilt part of the F positive by a small amount, and piston another part by a small amount.  Then we should be able to directly check at the simulated wavefront that we get results as expected for which side of the tilted segment moves in the same direction as a positive piston.

ivalaginja:
Oh man I just forgot that I am talking about the simulator here.... yeah that's the first thing I did, smaller motions so that the segments don't phase wrap. Easy enough. Thanks!

ivalaginja:
I double-checked the tip/tilt flip, and the sign on the PTT command. This should be ready.

catkit/emulators/iris_ao_controller.py
```diff
@@ -48,7 +48,13 @@ def convert_command_to_poppy_surface(dm_command_dict):
...
         # Setting the simulated IrisAO means setting each actuator individually
         for seg, values in convert_command_to_poppy_surface(new_surface).items():
-            self.set_actuator(seg-1, values[0] * u.um, values[1] * u.mrad, values[2] * u.mrad)  # TODO: double-check the -1 here, meant to correct for different segment names
+            # The IrisAO hardware component and the poppy hex DM object have flipped
+            # x/y coordinate axes, so we need to feed what is called "tilt" on the IrisAO
+            # into the "tip" argument from poppy and vice versa.
+            piston = values[0] * u.um
+            tip = values[2] * u.mrad
+            tilt = values[1] * u.mrad
+            self.set_actuator(seg-1, piston, tip, tilt)  # TODO: double-check the -1 here, meant to correct for different segment names
```
mperrin, line 57: Is the "# TODO: double check" comment needed any longer? Since you've tested this works and we're commanding the correct segments, seems to me probably not.

mperrin (approved):
Looks good to me

catkit/emulators/iris_ao_controller.py
ivalaginja: Yes it works, but I had literally just thrown the -1 in there when I got an error initially, and didn't think about it further as to why I needed to add it to make it work. I can remove it if you want? But I will keep a personal note to come back here to see what exactly is happening. Let me know.

mperrin: if you still don't understand it then ok leave the comment in for now.   But surely this is just a different convention for 0-based one 1-based indices. Poppy numbers segments starting from 0. I infer the Iris code does not, and starts from 1 instead. Right?

jamienoss (approved):

catkit/emulators/iris_ao_controller.py
ivalaginja: Yes, it will be something like this. The reason I want to double-check is that to my understanding, the commands are already taking care of that. So I just want to make sure it is really this.

ivalaginja:
The test seems to fail because one of the tests assumes a worse PSF when adding 0.25 microns of piston to one of the segments. Now that we start off with a relaxed IrisAO instead of a physically flat one, this is not necessarily true anymore.
I will fix this by flattening the IrisAO in this test first, as soon as I get to it.

PR #162, opened 2020-11-20
CATKIT-57: Enable up-down flip in ZWO camera

ivalaginja:
[CATKIT-57](https://jira.stsci.edu/browse/CATKIT-57)


ivalaginja:
@jamienoss correctly pointed out that this can be achieved with a 180 degree rotation + flip left-right, so this is not needed.

PR #170, opened 2020-12-14
HICAT-979 remove testbed_state refs from hardware code

jamienoss:
### Downstream PRs:
 * https://github.com/spacetelescope/jost-package/pull/29
 * https://github.com/spacetelescope/hicat-package/pull/447

### TODO:
 - [x] Test on hardware.

### DIFF
--- a/catkit/hardware/boston/BostonDmController.py
+++ b/catkit/hardware/boston/BostonDmController.py
@@ -3,7 +3,6 @@ import sys
 
 import numpy as np
 
-from catkit.hardware import testbed_state
 from catkit.interfaces.DeformableMirrorController import DeformableMirrorController
 
 # BMC is Boston's library and it only works on windows.
@@ -32,6 +31,8 @@ class BostonDmController(DeformableMirrorController):
         # Create class attributes for storing individual DM commands.
         self.dm1_command = None
         self.dm2_command = None
+        self.dm1_command_object = None
+        self.dm2_command_object = None
         self.serial_num = serial_number
         self.command_length = command_length
         self.dac_bit_width = dac_bit_width
@@ -90,8 +91,6 @@ class BostonDmController(DeformableMirrorController):
                 self.instrument.close_dm()
         finally:
             self.instrument = None
-            # Update testbed_state.
-            self.__close_dm_controller_testbed_state()
 
     def apply_shape_to_both(self, dm1_command_object, dm2_command_object):
         """Combines both commands and sends to the controller to produce a shape on each DM."""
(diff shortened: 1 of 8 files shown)


jamienoss:
@kjbrooks You're auto marked for review due to the changes to the iris code, however, you don't need to review this, and I'll merge once tests are complete etc.

I took a look at jost and ``testbed_state.iris_command_object`` isn't accessed from anywhere so we should be all good there.

There is, however, a change needed due to that of the ``ZwoCamera`` for which I've opened https://github.com/spacetelescope/jost-package/pull/29 for when you're ready to review and merge etc.

There are still a load of refs to ``testbed_state`` attributes for some of the sim classes in jost, however, they too aren't accessed from anywhere (other than where written to) so these can get nuked or not - it doesn't matter. Eventually, those remaining sim classes will use their emulated versions, once they're written, and can just be nuked then.

Let me know if you have any concerns.

jamienoss:
Note: The downstream hicat tests will fail.

catkit/hardware/iris_ao/iris_ao_controller.py
```diff
@@ -117,14 +113,12 @@ def zero(self, return_zeros=False):
...
         """
         zero_list = util.create_zero_list(util.iris_num_segments(self.config_id))
-        zeros = util.create_dict_from_list(zero_list, util.iris_pupil_naming(self.config_id))
-        self.send_data(zeros)
-
-        # Update the testbed state
-        self._update_iris_state(zeros)
+        dm_shape = SegmentedDmCommand(dm_config_id=self.config_id)
+        dm_shape.read_initial_command(zero_list)
+        self.apply_shape(dm_shape)
```
jamienoss, line 118: I need to test this on hardware.

@kjbrooks, @ivalaginja does this at least look correct, I wasn't sure?

ivalaginja: Yes, this looks fine to me (as long as `self` is an instance of some sort of iris ao object).

catkit/hardware/boston/BostonDmController.py
```diff
@@ -32,6 +31,8 @@ def initialize(self, serial_number, command_length, dac_bit_width):
         # Create class attributes for storing individual DM commands.
         self.dm1_command = None
         self.dm2_command = None
+        self.dm1_command_object = None
+        self.dm2_command_object = None
```
jamienoss, line 35: These aren't actually referenced, but I added them here for possible future use (they could equally be nuked 🤷 ).

catkit/hardware/boston/BostonDmController.py
```diff
@@ -32,6 +31,8 @@ def initialize(self, serial_number, command_length, dac_bit_width):
         # Create class attributes for storing individual DM commands.
         self.dm1_command = None
         self.dm2_command = None
```
jamienoss, line 33: These should be replaced by
```python
@property
def dm1_command(self):
  return self.dm1_command_object.to_command()

@property
def dm2_command(self):
  return self.dm2_command_object.to_command()
```

Once ``_open()`` calls ``apply_shape_to_both()`` instead of ``send_data()`` and there's a way for the controller to know the number of actuators (could prob do something with the ``command_length`` but that seemed yucky).

PR #181, opened 2021-01-12
CATKIT-61 Raise when Newport picomotor fails to move

jamienoss:
- [ ] PUNT Increase default of ``sleep_per_step``? 🤔  🤷 
- [x] Tested on hardware.

### DIFF
--- a/catkit/hardware/newport/NewportPicomotorController.py
+++ b/catkit/hardware/newport/NewportPicomotorController.py
@@ -195,7 +195,9 @@ class NewportPicomotorController(MotorController2):
         set_value = float(self._send_message(get_message, 'get')) - float(initial_value)
         
         if float(set_value) != value:
-            self.log.error(f'Something is wrong, {set_value} != {value}') 
+            error_msg = f"Newport Pico Motor failed to move as {set_value} != {value}. Try increasing sleep duration."
+            self.log.error(error_msg)
+            raise RuntimeError(error_msg)
          
         self.log.info(f'Command sent. Action : {cmd_key}. Axis : {axis}. Value : {value}')
 


catkit/hardware/newport/NewportPicomotorController.py
```diff
@@ -195,7 +195,9 @@ def command(self, cmd_key, axis, value):
         set_value = float(self._send_message(get_message, 'get')) - float(initial_value)
         
         if float(set_value) != value:
-            self.log.error(f'Something is wrong, {set_value} != {value}') 
+            error_msg = f"{self.__qualname__} failed to move as {set_value} != {value}. Try increasing sleep duration."
+            self.log.error(error_msg)
```
jamienoss, line 199: This device has its own log so I'm just rolling with this idiom.

catkit/hardware/newport/NewportPicomotorController.py
```diff
@@ -195,7 +195,9 @@ def command(self, cmd_key, axis, value):
         set_value = float(self._send_message(get_message, 'get')) - float(initial_value)
         
         if float(set_value) != value:
-            self.log.error(f'Something is wrong, {set_value} != {value}')
```
jamienoss, line 198: I encountered this error a handful of times when testing https://github.com/spacetelescope/hicat-package/pull/310 during the power and auto compensation phase of the npoint controller.

Doubling ``sleep_per_step`` from ``0.0005`` to ``0.001`` seemed to fix this. I'm not sure though if we'll need to make this a permanent change or whether it's indicative of dying hardware 😬 .

jamienoss: @RemiSoummer Have you ever seen this before?

PR #187, opened 2021-01-27
CATKIT-68: Fix IrisAO tip/tilt command signs

ivalaginja:
As reported in https://github.com/spacetelescope/hicat-package/pull/482, the IrisAO tip/tilt commands currently have the wrong sign, which I fix in this PR.

[CATKIT-68](https://jira.stsci.edu/browse/CATKIT-68)

### DIFF
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -51,8 +51,8 @@ class PoppyIrisAODM(poppy.dms.HexSegmentedDeformableMirror):
             # x/y coordinate axes, so we need to feed what is called "tilt" on the IrisAO
             # into the "tip" argument from poppy and vice versa.
             piston = values[0] * u.um
-            tip = values[2] * u.mrad
-            tilt = values[1] * u.mrad
+            tip = -values[2] * u.mrad    # negative sign matches the hardware behavior
+            tilt = -values[1] * u.mrad   # negative sign matches the hardware behavior
             self.set_actuator(seg-1, piston, tip, tilt)    # offset by -1 for 0-based vs 1-based segment indices; see PR #147
 
     @staticmethod
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -9,6 +9,7 @@ import catkit.hardware.iris_ao.segmented_dm_command as segmented_dm_command
 import catkit.hardware.iris_ao.util
 from catkit.interfaces.Instrument import SimInstrument
 import catkit.util
+from packaging.version import Version
 
 
 class PoppyIrisAODM(poppy.dms.HexSegmentedDeformableMirror):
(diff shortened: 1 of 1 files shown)


ivalaginja:
@kjbrooks does this have an impact on JOST?

jamienoss (commented):
😬  I'm ok for us to merge this as perhaps a temp fix, however, I would have thought that this was perhaps fixing only the symptom and not the cause. Isn't this sort of change indicative of an issue somewhere else?

This would suggest that the hardware does this implicitly, or our C++ code, or that the commanding is wrong, or just an issue in the sim.

I can merge without approval but we'll need to wait for a response from @kjbrooks.

kjbrooks:
This is the emulator so no impact on JOST, for right now. 

I am a little confused on one thing though: @ivalaginja have you compared this to the psf display in that exists in the hardware/iris_ao/segmented_dm_command? I would expect the emulator to match that (and likely replace it some day) so if it's not matching, I find that pretty odd. They should be doing the same thing, no?

mperrin:
Hi all. This is related to the work on sign conventions for WFE. Jamie’s right that it is indicative of an error somewhere else, or rather an inconsistency in sign conventions. As discussed elsewhere (in emails and poppy PR https://github.com/spacetelescope/poppy/pull/384), the sign conventions in Goodman’s Fourier Optics, on which poppy was originally based, is different than the sign conventions used by Wyantt and his various students and companies in Arizona, including 4D.

That poppy PR rationalizes all this and makes it fully consistent. One of the effects is a flip in the effect of tilt signs. That PR is near completion but I’ve been holding off in part because I hadn’t yet had a chance to work though all the implications and possible impacts on hicat. So this is good to see. 

My suggestion here for this PR is to check the `poppy.__version__` value, and if it’s less than 1.0 apply this sign flip, otherwise don’t. Or, alternatively, just hold off on this if possible for a brief period (days? Next week at the latest) and it should not be needed on poppy 1.0+

ivalaginja:
I am glad this makes sense in the context of those poppy changes. I see no reason why we would need this PR merged if the poppy fix is around the corner, even if it still takes a week or two. There is nobody using the tip/tilt controls on HiCAT IrisAO for now, so it's not breaking anything. I would suggest to keep this PR open as a reminder, and once we confirm everything is in order after the poppy update, we can simply close it. Does that sound ok?
@kjbrooks I haven't done that yet, but it's on my list of things to add to the orientation tests over in HiCAT.

jamienoss:
Let's bump this back to a draft and I'll block the approval just in case someone accidentally clicks on things (not that it would be a big deal).

jamienoss (changes requested):
See https://github.com/spacetelescope/catkit/pull/187#issuecomment-768427092

mperrin:
Upon further thought on this, I'd like us to proceed with my suggestion above (apply this flip depending on poppy `__version__`). This is going to make the catkit/hicat behavior for this independent of poppy version, which is going to make it easier to roll out this back-incompatible change in poppy without unintended disruptions in downstream packages or a need to closely coordinate dependency updates everywhere all at once. (The poppy fix is indeed around the corner, but it's gotten constrained in terms of when I can actually roll that out given multiple dependencies in JWST ops rehearsals as well as lab operations.)

I pushed a commit that adds the necessary version check.

jamienoss (approved):

PR #70, opened 2020-02-21
HICAT-714 Direct call utilization IrisAO API

jamienoss:
Resolves HICAT-714
    
Signed-off-by: James Noss <jnoss@stsci.edu>

PR #188, opened 2021-01-27
HICAT-294 Interface with IrisAO C/C++ API

jamienoss:
Resolves [HICAT-294](https://jira.stsci.edu/browse/HICAT-294) and relates to [HICAT-714](https://jira.stsci.edu/browse/HICAT-714).

Moved from #70

- [ ] Tested on hicta1 with spare.
- [ ] Tested on hicta2.
- [ ] Downstream hicat-package changes.
- [ ] Downstream jost-package changes.

https://github.com/spacetelescope/iris-ao
Turns out that we had the 64b lib all along 😬  😆  

Loading the following dll works. I haven't tested anything beyond that yet. https://github.com/spacetelescope/iris-ao/blob/vendor-installation-disk-v1.0.2.5a/Iris%20AO%20Installation%20Disk%20v1.0.2.5a/API/C-C%2B%2B%20Source%20Code%20%26%20Libraries/x64/IrisAO.Devices.dll

### DIFF
--- a/catkit/hardware/iris_ao/iris_ao_controller.py
+++ b/catkit/hardware/iris_ao/iris_ao_controller.py
@@ -1,13 +1,12 @@
 """Interface for IrisAO segmented deformable mirror controller."""
 
 import os
-import signal
-import subprocess
 import time
+
+import IrisAO_Python
 
 from catkit.interfaces.DeformableMirrorController import DeformableMirrorController
 from catkit.hardware.iris_ao.segmented_dm_command import SegmentedDmCommand
-
 from catkit.hardware.iris_ao import util
 
 
@@ -29,7 +28,7 @@ class IrisAoDmController(DeformableMirrorController):
     https://github.com/spacetelescope/catkit/pull/71#discussion_r466536405
     """
 
-    instrument_lib = subprocess
+    instrument_lib = IrisAO_Python
 
     def initialize(self,
                    mirror_serial,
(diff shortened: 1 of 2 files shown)


jamienoss:
@ivalaginja & @kjbrooks If we're no longer using the C++ broker, do we still want to write out the ``PTT.ini`` file?

ivalaginja:
@jamienoss no, we have functions to save used commands explicitly and without being overwritten, so the `PTT.ini` can be ditched. @kjbrooks do you agree?

jamienoss:
The use of "custom" in, e.g., ``custom_flat_filepath`` passed into the init is confusing since the .mcf (manufactures config file) is custom to the device.

Can we change this to ``user_flat_filepath``? This the path for ``CustomFlat.ini``, which we're now free to name arbitrarily.

Actually, this is not even relevant for this PR anymore - the dm is just zeroed (.mcf applied) and not truly flattened when opened.

ivalaginja:
> The use of "custom" in, e.g., `custom_flat_filepath` passed into the init is confusing since the .mcf (manufactures config file) is custom to the device.
> 
> Can we change this to `user_flat_filepath`? This the path for `CustomFlat.ini`, which we're now free to name arbitrarily.
> 
> Actually, this is not even relevant for this PR anymore - the dm is just zeroed (.mcf applied) and not truly flattened when opened.

@jamienoss I agree that the current naming can be confusing and have no objections to renaming it with your suggestion, unless it is not relevant, as you say.

kjbrooks:
@ivalaginja @jamienoss Yes, I agree on the PTT files. If you are also planning on getting ride of the ability to _read_ in the files, let me know so I can add that ability on the JOST side since we still have some of those.

ivalaginja:
@kjbrooks I believe he only meant the ConfigPTT.ini file in particular, the one that is written to disk in order for the C++ code to send it to the hardware. I would definitely keep the general ability to read ini command files. Or did I misunderstand @jamienoss ?

catkit/hardware/iris_ao/iris_ao_controller.py
```diff
@@ -1,106 +1,170 @@
...
+        # void* MirrorRelease (void* mirror);
+        self._MirrorRelease = self.dll.MirrorConnect
+        self._MirrorRelease.restype = ctypes.c_void_p
+        self._MirrorRelease.argtypes = [ctypes.c_void_p]
+
+    def MirrorCommand(self, mirror):
+        if self.positions_initiated:
+            return self._MirrorConnect(mirror, self.instrument_lib.MirrorSendSettings)
+        else:
+            # TODO: What's the diff?
```
jamienoss: It's possible that the difference is that ``MirrorInitSettings`` is the command to apply the .mcf flat. In which case, we might want to expose this, though, I'm not sure of the implications.

All of the C++ examples use ``MirrorInitSettings`` first and only once, and then ``MirrorSendSettings`` for all subsequent calls.

catkit/hardware/iris_ao/iris_ao_controller.py
```diff
@@ -123,23 +187,10 @@ def zero(self, return_zeros=False):
...
-            # However, the above no longer seems to work, see HICAT-948 & HICAT-947.
-            self.instrument.stdin.write(b'quit\n')
-            try:
-                self.instrument.stdin.flush()
-            except Exception:
-                pass
+            self.instrument_lib.MirrorRelease(ctypes.byref(self.instrument))
         finally:
             self.instrument = None
+            self._close_iris_controller_testbed_state()
```
jamienoss: and what are you doing here...?

catkit/hardware/iris_ao/iris_ao_controller.py
jamienoss: Manual says it's ok and lists the following example.
```C++
//This following code example demonstrates how to flatten the mirror segments:
// connect to the specified mirror device
MirrorHandle mirror = MirrorConnect(“FSC37-01-11-0613", “08130002", false);
// flatten all of the mirror segments
MirrorCommand(mirror, MirrorInitSettings);
// release the specified
MirrorRelease(mirror);
```

catkit/hardware/iris_ao/iris_ao_controller.py
```diff
@@ -1,109 +1,167 @@
...
+        self.mcf_filepath = mcf_filepath
+        self.dcf_filepath = dcf_filepath
+
+        # Check config files exist.
+        self.validate_config_files()
+
+        # The IrisAO driver expects the .dcf and .mcf config files in the working dir.
+        # Neither adding them to PATH nor PYTHONPATH worked.
+        # This is a workaround.
+        # These files remain copied until their refs go out of scope, i.e., when this class instance drops out of scope.
```
jamienoss, line 100: This is an old comment from when we tried using the vendor's python API.

It's more likely that these need to go in the same dir as the dll.

PR #189, opened 2021-02-01
HICAT-1014 Improved IrisAO subprocess error handling 

jamienoss:
- [x] Update emulator.

 Changes due to https://github.com/spacetelescope/hicat-package/pull/490

What we want:

* To read ``stdout`` and ``stderr``, however, we need to do so with a timeout for when there's nothing to read otherwise it will deadlock waiting for data that will never come.

What doesn't work:

 * ``subprocess.communicate()``, this is a single-use func and closes everything upon return.
 * `` select.select()``, doesn't work on Windows for pipes (I even tried just in case). 😭 

``select.select()`` seems to be the only python way to do this, which means we might have to abandon this.

The C++ https://github.com/spacetelescope/hicat-package/pull/490 removes all writes to ``stdout`` other than errors and "success" messages. Code here now reads these "success" messages. There is still scope for the deadlock of an indefinite read, however, I guess this is better than nothing. 🤷

### DIFF
--- a/catkit/hardware/iris_ao/iris_ao_controller.py
+++ b/catkit/hardware/iris_ao/iris_ao_controller.py
@@ -69,6 +69,16 @@ class IrisAoDmController(DeformableMirrorController):
         # Where to write ConfigPTT.ini file that gets read by the C++ code
         self.filename_ptt_dm = filename_ptt_dm
 
+    def _communicate(self, input):
+        try:
+            stdout, stderr = self.instrument.stdin.communicate(input=input, timeout=1)
+        except TimeoutExpired:
+            pass
+        if stdout:
+            self.log.info(stdout)
+        if stderr:
+            raise RuntimeError(stderr)
+
     def send_data(self, data):
         """
         To send data to the IrisAO, you must write to the ConfigPTT.ini file
@@ -85,9 +95,7 @@ class IrisAoDmController(DeformableMirrorController):
                        driver_serial=self.driver_serial)
 
         # Apply the written .ini file to DM
-        self.instrument.stdin.write(b'config\n')
-        self.instrument.stdin.flush()
-        self.raise_on_error()
+        self._communicate(input=b'config\n')
 
     def _open(self):
         """Open a connection to the IrisAO"""
@@ -109,11 +117,6 @@ class IrisAoDmController(DeformableMirrorController):
         self.zero()
 
         return self.instrument
-
-    def raise_on_error(self):
-        error = self.instrument.stderr.read()
-        if error:
-            raise RuntimeError(error)
 
     def zero(self, return_zeros=False):
         """Put zeros on the DM. This does not correspond to a flattened DM.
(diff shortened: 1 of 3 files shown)


catkit/hardware/iris_ao/iris_ao_controller.py
```diff
@@ -69,6 +69,32 @@ def initialize(self,
...
+            raise RuntimeError(f"{self.config_id} unexpectedly closed.") from error
+
+        # Poll in case it died (though the above flush would most likely catch this).
+        #if self.instrument.poll() is not None:
+        #    self.instrument = None  # If it died, there's nothing to close, so don't.
+        #    raise RuntimeError(f"{self.config_id} unexpectedly closed: ('{self.instrument.returncode}').")
+
+        # Read response.
+        self.log.info(f"{self.config_id}: Waiting for response...")
+        resp = self.instrument.stdout.readline().decode()
```
jamienoss, line 93: This could be an issue as it could block indefinitely. This read is not compatible with the existing builds of the C++ code in hicat-package.  Changes from https://github.com/spacetelescope/hicat-package/pull/490 requires rebuilds on hicat1, hiact2, & jost.

jamienoss:
@kjbrooks What version of catkit is JOST currently using? I tried looking in the repo but couldn't see an obvious reference.

kjbrooks:
@jamienoss JOST is using catkit v 0.24.0. Yes, there is no reference anywhere, but I will add it to the readme.

jamienoss:
@kjbrooks ok cool, so this PR will yield v0.26.0. Do you have a goal to use this soon or are you ok on v0.24.0 for a bit? I ask, because if you're ok, I can merge this and do the jost rebuild at a later date.

kjbrooks:
@jamienoss I am okay on 0.24.0 for a while. I am not doing much work on JOST for the time being so the rebuild can wait.

catkit/hardware/iris_ao/iris_ao_controller.py
jamienoss: @ivalaginja If the driver code blocks when talking to the dm such that it doesn't return until it's actually ready, this read will also block and wait until the driver func call returns, then our C++ writes to ``stdout``, which we read thus syncing Python and the DM.

If it was a sync issue of this nature that you were having with iterative dm communications, this should fix that problem. But that's quite a few ifs.

jamienoss: I've added a hardware test for this [here](https://github.com/spacetelescope/hicat-package/blob/119a0e072fc4eebb2436ee809391cdd1b8ed6943/hicat/hardware/iris_ao/tests/hardware_speed.py).

jamienoss: Though, that test only pokes single segments which may be faster than having to do all of them per iter.

ivalaginja: Many ifs, but might rule some things out. Excellent that you put this in place, thanks!
I don't think doing a single segment is much faster. The IrisAO is both extremely repeatable and extremely fast in its commands, and on JOST we observed the deadness of connection in iterations that did something that's almost exactly what you wrote up in the test, just tilting a single segment more and more, iteratively. We'll  have to try and see.

jamienoss: See https://github.com/spacetelescope/hicat-package/pull/490#issuecomment-772775083, this PR solves your problem as I can reproduce your stall without the added ``readline()`` wait.

PR #191, opened 2021-02-02
Fix typo in PTT111 figure

kjbrooks:
I had mislabelled on of the segments in this figure. This new figured has fixed that typo.


jamienoss:
@kjbrooks The tests don't cover this, right, so should I just merge?

kjbrooks:
@jamienoss that is correct. Yup, that would be great!

PR #194, opened 2021-02-09
CATKIT-76 Make DeviceCacheEnum mutable and add ImmutableDeviceCacheEnum

jamienoss:

### DIFF
--- a/catkit/testbed/caching.py
+++ b/catkit/testbed/caching.py
@@ -220,6 +220,7 @@ class DeviceCacheEnum(Enum):
     def __init__(self, description, config_id):
         self.description = description
         self.config_id = config_id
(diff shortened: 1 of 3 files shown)


catkit/testbed/caching.py
```diff
@@ -236,6 +237,22 @@ def __getattr__(self, item):
         device = devices[member]
         return device.__getattribute__(item)
 
+    def __setattr__(self, name, value):
+        if "_DeviceCacheEnum__object_exists" in self.__dict__:
+            config_id = object.__getattribute__(self, "config_id")
```
jamienoss, line 242: Copy and paste error, this is unnecessary and should just be ``config_id = self. config_id``.

PR #198, opened 2021-02-17
CATKIT-78 Remove testbed_state functionality and refs

jamienoss:
- [x] Check impact on JOST https://github.com/spacetelescope/jost-package/pull/32

This PR will break hicat-package without https://github.com/spacetelescope/hicat-package/pull/518.


jamienoss:
NOTE: Downstream hicat tests will fail.

PR #199, opened 2021-02-19
HICAT-1042: Enable video capture mode for ZWO cameras

ehpor:
Enables video capture mode for the ZWO cameras. This PR also fixes some minor outdated documentation related to the zwo camera capture functions. An accompanying hicat-package PR fixes simulator calls.

Timing (for 1ms exposures on the PR camera):
* with snapshot mode: 500 images took 62.5 secs
* with video mode: 500 images took 5.47 secs

Some remaining questions:
* How should exception handling should be done. Should I convert ZWO exceptions into runtime errors?
* How should timeouts be handled? How long should we wait? Timeouts can be infinite, but I set it to the SDK recommended timeout.

### DIFF
--- a/catkit/emulators/ZwoCamera.py
+++ b/catkit/emulators/ZwoCamera.py
@@ -53,7 +53,7 @@ class ZwoEmulator(ZwoASI):
         return self
 
     def get_controls(self):
-        # only used for oepn behavior to 
+        # only used for oepn behavior to
         # get / set control values to default on open
         # needs to play nicely with calls to set_controls
         # this phony dict is set to have *some* accessible value (None) for
@@ -65,6 +65,9 @@ class ZwoEmulator(ZwoASI):
         if value is not None and not isinstance(value, accepted_types):
             raise ValueError(f"Expected type {accepted_types} got '{type(value)}'")
         self.control_values[control_type] = value
+
+    def start_video_capture(self):
+        pass
 
     def stop_video_capture(self):
         pass
@@ -79,6 +82,10 @@ class ZwoEmulator(ZwoASI):
     def capture(self, initial_sleep=0.01, poll=0.01, buffer=None, filename=None):
         pass
 
+    @abc.abstractmethod
+    def capture_video_frame(self, buffer=None, filename=None, timeout=None):
+        return self.capture(buffer=buffer, filename=filename)
+
     def close(self):
         pass
 
@@ -88,7 +95,7 @@ class ZwoEmulator(ZwoASI):
         # I *think* this can just return a dict with milquetoast values for this
         # took said milquetoast values from the __setup_control_values sim
         # function here
-        return {'MaxWidth': 4096, 'MaxHeight': 4096} 
+        return {'MaxWidth': 4096, 'MaxHeight': 4096}
         pass
 
     def set_id(self, id, id_str):
@@ -97,8 +104,8 @@ class ZwoEmulator(ZwoASI):
     def set_roi(self, start_x=None, start_y=None, width=None, height=None, bins=None, image_type=None):
         # sets region of interest
         # purpose : set_roi_format, set_roi_start_position
-        # set_roi_format --> _set_roi_format : 
-        # check for all the errors 
+        # set_roi_format --> _set_roi_format :
+        # check for all the errors
         # runs zwolib.ASISetROIFormat(id_, width, height, bins, image_type)
         # set_roi_start_position --> _set_start_position :
         # runs zwolib.ASISetStartPos(id_, start_x, start_y)
--- a/catkit/emulators/ZwoCamera.py
+++ b/catkit/emulators/ZwoCamera.py
@@ -82,7 +82,6 @@ class ZwoEmulator(ZwoASI):
     def capture(self, initial_sleep=0.01, poll=0.01, buffer=None, filename=None):
         pass
 
-    @abc.abstractmethod
     def capture_video_frame(self, buffer=None, filename=None, timeout=None):
         return self.capture(buffer=buffer, filename=filename)
 
--- a/catkit/hardware/zwo/ZwoCamera.py
+++ b/catkit/hardware/zwo/ZwoCamera.py
@@ -183,6 +183,8 @@ class ZwoCamera(Camera):
 
                 images.append(img.astype(np.dtype(np.float32)))
         finally:
+            # Stop exposures. The stop_exposure() might not be necessary, but there's no
+            # harm in calling it anyway.
             self.instrument.stop_video_capture()
             self.instrument.stop_exposure()
 
@@ -262,7 +264,7 @@ class ZwoCamera(Camera):
         :param gain: Gain of ZWO camera (volts).
         :param full_image: Boolean for whether to take a full image.
         :param bins: Integer value for number of bins.
-        :param use_video_capture_mode: Boolean for whether to use video capture or snapshot mode.
+        :param use_video_capture_mode: Boolean for whether to use video capture or snapshot mode. Default is False.
         :return: Two parameters: Image list (numpy data or paths), Metadata list of MetaDataEntry objects.
         """
 
--- a/catkit/hardware/zwo/ZwoCamera.py
+++ b/catkit/hardware/zwo/ZwoCamera.py
@@ -250,7 +250,7 @@ class ZwoCamera(Camera):
     def just_take_exposures(self, exposure_time, num_exposures,
                             extra_metadata=None,
                             subarray_x=None, subarray_y=None, width=None, height=None, gain=None, full_image=None,
-                            bins=None, use_video_capture_mode=None):
+                            bins=None, use_video_capture_mode=False):
         """
         Low level method to take exposures using a Zwo camera. By default keeps image data in.
 
@@ -267,10 +267,6 @@ class ZwoCamera(Camera):
         :param use_video_capture_mode: Boolean for whether to use video capture or snapshot mode. Default is False.
         :return: Two parameters: Image list (numpy data or paths), Metadata list of MetaDataEntry objects.
         """
-
-        # Set default for using video capture mode.
-        if use_video_capture_mode is None:
-            use_video_capture_mode = False
 
         # Convert exposure time to contain units if not already a Pint quantity.
         # if not isintance(quantity):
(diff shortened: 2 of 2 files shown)


catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -147,14 +147,71 @@ def __capture_and_orient(self, initial_sleep, theta, fliplr):
...
+        self.instrument.start_video_capture()
+
+        images = []
+        for i in range(num_exposures):
+            img = self.instrument.capture_video_frame(timeout=timeout.to(units.millisecond).magnitude)
+
+            images.append(img.astype(np.dtype(np.float32)))
+
+        self.instrument.stop_video_capture()
+        self.instrument.stop_exposure()
```
jamienoss: [revision?] I think this needs to be:
```python
self.instrument.start_video_capture()
try:
  ...
finally:
  self.instrument.stop_video_capture()
  self.instrument.stop_exposure()
```

catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -147,14 +147,71 @@ def __capture_and_orient(self, initial_sleep, theta, fliplr):
...
+        Returns
+        -------
+        images : list of (np.array of floats)
+            All captured images in a list.
+        """
+        self.instrument.start_video_capture()
+
+        images = []
+        for i in range(num_exposures):
+            img = self.instrument.capture_video_frame(timeout=timeout.to(units.millisecond).magnitude)
```
jamienoss: [revision] Let's hoist the ``timeout`` above the loop, please.

catkit/hardware/zwo/ZwoCamera.py
```diff
@@ -202,9 +259,14 @@ def just_take_exposures(self, exposure_time, num_exposures,
         :param gain: Gain of ZWO camera (volts).
         :param full_image: Boolean for whether to take a full image.
         :param bins: Integer value for number of bins.
+        :param use_video_capture_mode: Boolean for whether to use video capture or snapshot mode.
         :return: Two parameters: Image list (numpy data or paths), Metadata list of MetaDataEntry objects.
         """
 
+        # Set default for using video capture mode.
+        if use_video_capture_mode is None:
+            use_video_capture_mode = False
```
jamienoss: [revision] Let's make the actual default ``False``, please, and let ``None := False``, which ``if use_video_capture_mode:`` does.

catkit/hardware/zwo/ZwoCamera.py
jamienoss: What happens when it times out? Does it raise or...?

catkit/emulators/ZwoCamera.py
```diff
@@ -79,6 +82,10 @@ def set_image_type(self, image_type):
     def capture(self, initial_sleep=0.01, poll=0.01, buffer=None, filename=None):
         pass
 
+    @abc.abstractmethod
+    def capture_video_frame(self, buffer=None, filename=None, timeout=None):
+        return self.capture(buffer=buffer, filename=filename)
```
jamienoss, line 86: 👍

catkit/emulators/ZwoCamera.py
```diff
@@ -79,6 +82,10 @@ def set_image_type(self, image_type):
     def capture(self, initial_sleep=0.01, poll=0.01, buffer=None, filename=None):
         pass
 
+    @abc.abstractmethod
```
jamienoss: [revision] This needs to be removed. It doesn't make sense to be abstract _and_ have a def.

jamienoss (changes requested):

catkit/hardware/zwo/ZwoCamera.py
jamienoss: Why the ``self.instrument.stop_exposure()``?

catkit/emulators/ZwoCamera.py
ehpor: Agreed. I forgot to remove that. Sorry.

catkit/hardware/zwo/ZwoCamera.py
ehpor: It raises a ZWO_IOError (with code 11: timeout). We might want to convert this to a RuntimeError? We can also not set the timeout. Then it will just hang and wait forever on an image that may never get there. The SDK is unclear on if a ZWO_IOError (with code 5: camera removed) will be raised if the USB disconnects during an exposure. You might know from experience if this is the case; I'm unable to test this as I cannot jank the USB cable out while taking images.

catkit/hardware/zwo/ZwoCamera.py
ehpor: I did this to let catkit, and only catkit, define the default, and still be able to pass through a use_video_capture from hicat-package.

You're saying set the default to False, but then hicat-package, would set the default. Or do you mean, manually covert Nones to Falses still, but just set the default to False? That doesn't sound clean to me.

catkit/hardware/zwo/ZwoCamera.py
ehpor: Agreed.

catkit/hardware/zwo/ZwoCamera.py
ehpor: Agreed.

The SDK is unclear (worst SDK I've seen on such an expensive camera, btw) on if stop video capture actually stops the current exposure or not. As in the `_open()` function, this combo is used to halt any exposure, I just used the exact same function here. The API does not state either way, if it is for snapshot captures only, or also video capture mode exposures. It doesn't do any harm calling it anyway.

If you want me to, I can test if `stop_video_exposure()` just stops the next frame from starting, or if it also cancels the current exposure.

catkit/emulators/ZwoCamera.py
ehpor: Yep, you were right after all. This doesn't need to be in hicat-package.

catkit/hardware/zwo/ZwoCamera.py
jamienoss: >It raises a ZWO_IOError (with code 11: timeout). We might want to convert this to a RuntimeError? 

This is cool as is.

>We can also not set the timeout. Then it will just hang and wait forever on an image that may never get there. 

We certainly don't want to do that. The sooner something tanks out the better.

>The SDK is unclear on if a ZWO_IOError (with code 5: camera removed) will be raised if the USB disconnects during an exposure. You might know from experience if this is the case; I'm unable to test this as I cannot jank the USB cable out while taking images.

No experience there, sorry, but I don't think we care too much either, or even if we did we could do much about it anyhow.

catkit/hardware/zwo/ZwoCamera.py
jamienoss: I mean:
```python
 def just_take_exposures(self, ... , use_video_capture_mode=False)
```

> You're saying set the default to False, but then hicat-package, would set the default.

I don't really follow what you mean here?

jamienoss: Since you already define ``None := False``, they're logically the same thing, however, it's not obvious that this is actually the default as your dec is in the code and not the func sig.

catkit/hardware/zwo/ZwoCamera.py
jamienoss: The SDK is not provided by the vendor but an individual in the community, who no longer has time to work on it (they even asked me to at one point, but I neither have time).

Nah, leave it in then. Your comment here is good enough to cover it if we wanna know why in the future.

catkit/hardware/zwo/ZwoCamera.py
ehpor: Okay, so you meant the latter.

But then you have multiple places in which the default is listed (1. in the function signature, and 2. in the function body). Setting the first one to True later on, and forgetting to change the second, will mean that the default from the just_take_exposures() call from HicatZwoCamera (in hicat-package, which currently has None as its default), will still mean False. By your logic, I would want to change the default in HicatZwoCamera as well. That default is now None, and should be False/True to make clear what the default is. The same is true for the simulator version of ZwoCamera (in hicat-package simulator). That would mean that you would need to keep at least *four* places all over the codebase in sync with each other to have a consistent default. And missing to change only one, means that you'll have inconsistent behaviour.

Setting all function signature defaults to None ensures that this will never happen. There is only one place where the default is defined (in the function body), and the docstring tells you what this default is (I'll add that after we agree on something, I forgot to do that before asking for your review).

catkit/hardware/zwo/ZwoCamera.py
ehpor: My main worry is that, only very sometimes, the timeout may trigger too soon. It hasn't happened yet during testing, but I'm sure it'll happen at some point. Do we want to set the timeout to an insane value (say 10-30 seconds) to ensure that the timeout only triggers when something actually went wrong?

catkit/hardware/zwo/ZwoCamera.py
ehpor: Nah, I meant the ZWO C API documentation. The python SDK is doing the best he can with the limited information that the C API documentation provides. Link to the C API documentation: https://condorarraytelescope.org/static/pages/pdf/manuals/ZWO-0005.pdf.

catkit/hardware/zwo/ZwoCamera.py
jamienoss: We could do something like 10-30s, however, it really should be a function of the exposure time. We could, however, _add_ 60s (usual default timeout) to that derived from the exp-time, i.e., to account for when the exp-time _is_ 60s.

jamienoss: Some % would also work, i.e., timeout = f(exp-time)*0.2.

catkit/hardware/zwo/ZwoCamera.py
ehpor: Changed the default parameters to False instead of None.

ehpor:
Seems to still work on the bench. Ready for second review. Timeout is now: 2 * exposure time + 10 seconds. This keeps the timeout under 32767 milliseconds, in case the underlying value is a signed short. Ten seconds is more than enough time for us to say that the camera actually crashed.

jamienoss (approved):

PR #200, opened 2021-02-22
CATKIT-20 Add and use catkit.util.sleep()

jamienoss:


catkit/util.py
```diff
@@ -12,6 +12,15 @@
 from catkit.catkit_types import quantity
 
 
+simulation = False
```
jamienoss, line 15: This'll do for now...

PR #203, opened 2021-02-23
HICAT-1046 Store command state in Iris object

jamienoss:
This was already the case, so this actually only clears the command state upon error.


jamienoss:
@kjbrooks FYI This can just be blindly absorbed into JOST (it doesn't require any attention nor changes etc).

PR #204, opened 2021-02-24
CATKIT-80 Refactor MFF101 to use Instrument interface

jamienoss:
This is also:

 * Adds the emulator.
 * dedupes and refactors the code.
 * Store position state as ``current_position``
 * Allows ``move_to_position()`` to accept ``FlipMountPosition``
 * Adds ``move()`` as alias to ``move_to_position()``.


jamienoss:
This _will_ break the downstream hicat tests.

jamienoss:
@kjbrooks This doesn't impact JOST as it doesn't have any MFF101 (at least it's not using this code).

I've also added two new labels ``Impacts JOST`` and ``Does NOT impact JOST``.

PR #210, opened 2021-03-02
CATKIT-83 Poll IrisAO subprocess after sending commanding

jamienoss:
See https://github.com/spacetelescope/catkit/pull/189#discussion_r568275966 for details.

### DIFF
--- a/catkit/hardware/iris_ao/iris_ao_controller.py
+++ b/catkit/hardware/iris_ao/iris_ao_controller.py
@@ -85,10 +85,13 @@ class IrisAoDmController(DeformableMirrorController):
             self.instrument = None  # If it died, there's nothing to close, so don't.
             raise RuntimeError(f"{self.config_id} unexpectedly closed.") from error
 
-        # Poll in case it died (though the above flush would most likely catch this).
-        #if self.instrument.poll() is not None:
-        #    self.instrument = None  # If it died, there's nothing to close, so don't.
-        #    raise RuntimeError(f"{self.config_id} unexpectedly closed: ('{self.instrument.returncode}').")
+        # Poll in case it died after receiving the above commanding.
+        if self.instrument.poll() is not None:
+            self.instrument = None  # If it died, there's nothing to close, so don't.
+            raise RuntimeError(f"{self.config_id} unexpectedly closed: ('{self.instrument.returncode}').")
+
+        # WARNING!!! - even with the above poll() there's still scope for the process to have died in this gap.
+        # In which case, the below readline() will deadlock! See https://github.com/spacetelescope/catkit/pull/189.
 
         # Read response.
         self.log.info(f"{self.config_id}: Waiting for response...")


catkit/hardware/iris_ao/iris_ao_controller.py
```diff
@@ -85,10 +85,13 @@ def _communicate(self, input):
             self.instrument = None  # If it died, there's nothing to close, so don't.
             raise RuntimeError(f"{self.config_id} unexpectedly closed.") from error
```
jamienoss, line 87: A sleep in here would make this more robust and more likely for the poll to catch the terminated process, however, it would be at the obvious cost of slowing things down. It would prob have to be the order of 1s, so I don't suggest we do it.

catkit/hardware/iris_ao/iris_ao_controller.py
```diff
@@ -85,10 +85,13 @@ def _communicate(self, input):
...
-        #if self.instrument.poll() is not None:
-        #    self.instrument = None  # If it died, there's nothing to close, so don't.
-        #    raise RuntimeError(f"{self.config_id} unexpectedly closed: ('{self.instrument.returncode}').")
+        # Poll in case it died after receiving the above commanding.
+        if self.instrument.poll() is not None:
+            self.instrument = None  # If it died, there's nothing to close, so don't.
+            raise RuntimeError(f"{self.config_id} unexpectedly closed: ('{self.instrument.returncode}').")
+
+        # WARNING!!! - even with the above poll() there's still scope for the process to have died in this gap.
+        # In which case, the below readline() will deadlock! See https://github.com/spacetelescope/catkit/pull/189.
```
jamienoss, line 94: Really, I should check whether a terminated process before ``readline()`` will not cause the ``readline()`` to fail - TBH, I would have thought that it would in the same way that we see when this happens for a flush as th pipe to ``stdout`` gets closed. If it does, then the process issue happens during the ``readline()`` and this PR does nothing 🤷  (which is why I left it out in the first place - along with the fact that even if it did do something it's still racey).

kjbrooks:
@jamienoss To make sure I understand what this is fixing, is this the fix for when the IrisAO shape is not updated when you have loop? From your comments, it sounds like you aren't actually sure of this fix. Have you tested this fix on the hardware yet, or do you plan to when there is access to the testbed?

ivalaginja:
@kjbrooks it's a follow-up to my observation that the IrisAO hung itself up again in one of my pastis experiments:
https://stsci.slack.com/archives/GBCLQKHM3/p1614600404043800
This made the entire script stall because it was just hanging there, waiting for a response from the IrisAO.

It just stopped responding in a random iteration. We have seen this before, but Jamie added some fixes in #189 that made it look like it went away. It *mostly* works now, with occasional stalling situations like this, and this PR is attempting to improve on that.

ivalaginja:
One of the problems is that we can't really trigger the problem purposefully, at least I wouldn't know how. The last two times this happened was in the 70th, then in the 968th iteration of one and the same experiment (after restarting), so there seems to be nothing in particular that triggers it.

jamienoss:
@kjbrooks Hey, so there was an issue before where subsequent commands were being sent to the iris/C++ faster than it could deal with them. This was fixed by #189, however, our hands were tied, and that PR created its own problem, which @ivalaginja is now hitting.

It was fixed by adding the ``readline()`` that waits for a response from our C++ code before trying to send another command. However, if the C++ dies just before or during the ``readline()``, the whole thing locks up and hangs (deadlock) - it waits for something to read that will never come. What we want is ``readline(timeout=60)`` but that doesn't exist and workarounds to create this don't exist for Windows pipes. A soln. exists, it just means re-writing the C++ entirely or getting the direct python API to work.

I don't see any issue with this breaking anything, however, it may not actually fix the problem. It could mitigate it but the Gremlins will still be there, waiting for us...

I haven't tested this on hardware.

kjbrooks (approved):

kjbrooks:
Got it! Thanks for explaining the issue @jamienoss and @ivalaginja !

ehpor:
Doesn't threading provide a possible solution? Communicating with your subprocess pipes in a second thread and appending the read lines from the pipe into a list/queue. Then in your main thread, just poll that list/queue for new lines for at most `timeout` seconds with a small sleep in between polls, whenever you want to a read with timeout.

jamienoss:
@ehpor Maybe, but I'm not sure. I'm not sure of the nuances of python's pseudo-threading and whether it uses the low-level sleep primitives. Multiproc would def work, for example.

My concern would be whether the consumer thread (reading from the pipe) would actually sleep when not needed or whether it causes the GIL to context switch and thus yield a hit on perf throughout the entire lifetime of the Iris being "open", which would be the entire lifetime of an experiment run that uses it.

The alt. would be to spin-up the thread every time we comm with the Iris, however, I'm not sure of the overhead there, I'd guess that it might not be fast enough. Of course, it might be better than nothing, as I've said before.

At the end of the day, however, we should spend more time on what went wrong, the cause not the symptom, though a fix here might help us find this.

catkit/hardware/iris_ao/iris_ao_controller.py
jamienoss: @ivalaginja If/when bug-hunting I would add a ``time.sleep(1)`` here (maybe even 2s).

jamienoss:
I've merged this, if we want to do better we can do so in a future PR.

jamienoss:
@ivalaginja Also, not that hicat won't be using this ver of catkit so you'll have to checkout this commit c5628378ebb4882fedd9cd36577a5b96e609ae53

catkit/hardware/iris_ao/iris_ao_controller.py
ivalaginja: Right - I just wouldn't know how to hunt for bugs. The stalling seems to happen in random iterations of experiments that last 2+ hours 🤷

jamienoss: You could check the command file, whilst it's still hanging (or post exception), in case it was sent some bogus data. You could also try connecting to it from the GUI.

jamienoss: For the latter, you prob don't want the sleep, such that you can do it whilst it's still hanging.

You could take a look at the cams in sharpcap, might help to confirm that it actually died.

ivalaginja: Yeah I did check Sharpcap, it was just holding the DH. All DMs seemed to just keep holding their latest shape.

PR #212, opened 2021-03-05
Make video mode default for ZWO cameras

jamienoss:


ehpor (approved):
Seems good. I know you know this, but for documentation purposes: there are still two default parameters in hicat-package that need switching: in `hicat/hardware/testbed.py#L280` and `hicat/simulators/SimZwoCamera.py#L180`.

PR #216, opened 2021-03-12
Add reversed sign based on poppy version in piston too

ivalaginja:
Follow-up to #187 

If we assume poppy to be consistent within itself, which it is, we will also need a sign inversion in piston when passing IrisAO commands to the emulator, this PR adds that, in the same way like for the tip/tilt commands.

@ehpor was running dOTF experiments on HiCAT with which we could confirm this.


ivalaginja:
@ehpor I didn't want to forget about this so I simply opened the PR. Feel free to approve or change things whenever you know more from your experiments, although I am 90% sure we'll need this fix.

jamienoss (approved):

jamienoss:
We still want this, right?

mperrin:
yes, still wanted

jamienoss:
This was reverted in #231 and reopened in #232.

PR #218, opened 2021-03-22
CATKIT-91 Fix numpy size compatibility issue.

jamienoss:
Summary (ignoring chaos and stupidity):

numpy 1.20 added stuff to the ndarray C struct which changed its size. This results in a compatibility issue with some compiled code, e.g., like that in photutils, throwing the error:
```python
E   ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

1. catkit CI runs its tests and hicat's in separate CI jobs, however, the catkit-only tests were still using the hicat env which caused me to waste a day chasing my own tail.
2. photutils 1.1.0 (rel March 20) from conda-forge seems to be built against numpy 1.20 (which is our issue)
3. photutils 1.0.2 (rel Jan 20) was not and works. This is what we were using before when tests passed, along with numpy 1.19.2.
4. Both of these versions work when obtained from pip (even as bin wheels) so we're doing this hoping never to encounter this issue until we _need_ numpy >= 1.20.
5. Our numpy dep is now ``>=1.15.0,<1.20``.
6. Downstream catkit tests of hicat will fail until this is reflected in the hicat environment.yml also.
 
### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -10,7 +10,7 @@
 name: catkit
 dependencies:
   - astropy>=1.3
-  - numpy==1.16.6
+  - numpy==1.17.5
   - conda-forge::pyusb
   - flake8
   - h5py
--- a/environment.yml
+++ b/environment.yml
@@ -28,7 +28,6 @@ dependencies:
   - scipy
   - sphinx
   - pip:
-    - zwoasi>=0.0.21 # Requires additional manual install of driver(s) from https://astronomy-imaging-camera.com/software-drivers
-    - ftd2xx # Requires additional manual install of driver from https://www.ftdichip.com/Drivers/D2XX.htm
-    - hcipy
-    - git+https://github.com/spacetelescope/poppy.git@v0.9.2
+    - --no-binary
+    - --no-build-isolation
+    - -r file:requirements.txt
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - flake8
   - h5py
   - matplotlib
-  - conda-forge::photutils==0.7.2
+  - http://ssb.stsci.edu/astroconda::photutils==0.7.2
   - pip
   - psutil
   - pyserial
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - flake8
   - h5py
   - matplotlib
-  - conda-forge::photutils==1.0.2
+  - conda-forge::photutils==1.0.1
   - pip
   - psutil
   - pyserial
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - flake8
   - h5py
   - matplotlib
-  - conda-forge::photutils==1.0.1
+  - conda-forge::photutils==1.0.0
   - pip
   - psutil
   - pyserial
--- a/environment.yml
+++ b/environment.yml
@@ -9,6 +9,17 @@
 
 name: catkit
 dependencies:
+  - cryptography=3.3.1
+  - flake8=3.8.4
+  - hypothesis=6.6.1
+  - importlib-metadata=2.0.0
+  - importlib_metadata=2.0.0
+  - libtiff=4.1.0
+  - sphinx=3.5.2
+  - sqlite=3.33.0
+  - toml=0.10.1
+  - zipp=3.4.0
+  - zope.interface=5.2.0
   - conda-forge::astropy>=1.3
   - conda-forge::numpy>=1.15.0  # See CATKIT-91
   - conda-forge::numexpr  # See CATKIT-91
--- a/environment.yml
+++ b/environment.yml
@@ -28,6 +28,8 @@ dependencies:
   - scipy
   - sphinx
   - pip:
+    - --no-binary
+    - --no-build-isolation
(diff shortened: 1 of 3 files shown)


jamienoss:
Should test ``numpy>=1.20.0`` next.

jamienoss:
Tie-down photutils next... (tried this before but I think more than one thing is at play here).
See https://github.com/astropy/photutils/issues/1149

jamienoss:
Might also be worth trying ``numpy`` from ``conda-forge``, since ``photutils`` is, it may be built against that (but even build changes there could cause this, so how deep does the rabbit hole go).

Perhaps we just need to wait for an upstream fix... 🤞

jamienoss:
Follow the breadcrumbs:

1. https://stackoverflow.com/questions/66060487/valueerror-numpy-ndarray-size-changed-may-indicate-binary-incompatibility-exp
2. https://github.com/scikit-learn-contrib/hdbscan/issues/457#issuecomment-773671043
3. https://github.com/pypa/pip/issues/9542

requirements.txt
```diff
@@ -0,0 +1,5 @@
+--no-binary :all:
+zwoasi>=0.0.21 # Requires additional manual install of driver(s) from https://astronomy-imaging-camera.com/software-drivers
+ftd2xx # Requires additional manual install of driver from https://www.ftdichip.com/Drivers/D2XX.htm
+hcipy
+git+https://github.com/spacetelescope/poppy.git@v0.9.2
```
jamienoss: I don't really see how this relates to ``photutils``? Maybe we should just grab it from github.

jamienoss:
Is this the issue https://github.com/astropy/photutils/blob/main/pyproject.toml#L6? 

From what I've read the issue is caused by numpy >=1.20.0 but our tests have failed with numpy 1.19.2. Though it could be the issue that conda installs  1.19.2 but something else is using 1.20.x fr their build? But then we tried that by also using 1.20.x... ah, but then photutils wouldn't have been using 1.20.x...

jamienoss:
🤣 
```
Collecting git+https://github.com/astropy/photutils.git@1.0.2 (from -r file:requirements.txt (line 6))

  Cloning https://github.com/astropy/photutils.git (to revision 1.0.2) to c:\users\runner~1\appdata\local\temp\pip-req-build-roaj084y

  Installing build dependencies: started

  Installing build dependencies: still running...

  Installing build dependencies: still running...

  Installing build dependencies: still running...

  Installing build dependencies: finished with status 'error'  # <--------------


failed
Error: Process completed with exit code 1.
```

jamienoss:
https://github.com/numpy/numpy/releases/tag/v1.20.0
>C API changes
>#define NPY_SIZEOF_PYARRAYOBJECT (sizeof(PyArrayObject_fields))
since the size must not be considered a compile time constant: it will
change for different runtime versions of NumPy.
>
>The most likely relevant use are potential subclasses written in C which
will have to be recompiled and should be updated. Please see the
documentation for :cPyArrayObject{.interpreted-text role="type"} for
more details and contact the NumPy developers if you are affected by
this change.
>
>NumPy will attempt to give a graceful error but a program expecting a
fixed structure size may have undefined behaviour and likely crash.
>
>(gh-16938)

jamienoss:
numpy=1.16.6 worked but caused photutils=0.7.2.

environment.yml
```diff
@@ -21,7 +21,7 @@ dependencies:
   - zipp=3.4.0
   - zope.interface=5.2.0
   - conda-forge::astropy>=1.3
-  - conda-forge::numpy>=1.15.0  # See CATKIT-91
+  - conda-forge::numpy>=1.19.2  # See CATKIT-91
```
jamienoss: Whoops, I should have paid better attention from the beginning. I totally missed the ``>``. I thought I had double-checked that this used 1.19.2 but doing so again it didn't it used 0.20.1.

jamienoss: Actually, scratch that, ``origin/develop`` was using 1.19.2 as we can see [here](https://github.com/spacetelescope/catkit/runs/2176266394#step:5:503) but ``photutils=1.1.0`` ([here](https://github.com/spacetelescope/catkit/runs/2176266394#step:5:310)).

Using ``numpy=1.17.5`` also failed and picked up ``photutils=1.1.0``.

environment.yml
```diff
@@ -28,6 +27,7 @@ dependencies:
   - scipy
   - sphinx
   - pip:
+    - photutils
```
jamienoss: I expected this to fail, gonna need to bust out the ``requirements.txt`` again...

requirements.txt
```diff
@@ -0,0 +1,6 @@
+- --no-binary :all:
```
jamienoss: May need to also add ``--no-build-isolation``, but let's try this first as, after that, I'm not sure what to do but 😭

jamienoss:
ok, I think I have this solved now, finally... I've been chasing my own tail as the catkit tests still use the hicat env when they shouldn't be. Them not doing so (as I thought at least) was the reason I've been testing this here and not in hicat.

jamienoss:
https://github.com/spacetelescope/catkit/pull/218/commits/53ef507441e7aa4f0552bd1c954750cd53209572 worked for both catkit tests but failed the downstream hicat test but only for windows.

PR #227, opened 2021-04-26
HICAT-1082: Add Thorlabs power meter

ehpor:
This implements the Thorlabs Power Meter interface. This closely mimics the TSP01 interface, since the libraries are pretty much the same.

* [x] Test with power meter connected.
* [x] Add emulator ~~(may need to be in hicat-package instead).~~
* [x] Retesting on hardware, because a lot of the underlying code has changed in the meantime.

### DIFF
--- a/catkit/emulators/thorlabs/PM.py
+++ b/catkit/emulators/thorlabs/PM.py
@@ -1,0 +1,49 @@
+from abc import ABC, abstractmethod
+
+
+class PMEmulator(ABC):
+    """ Emulates the TLPM interface. """
+
+    # int TLPM_errorMessage(void *vi, int statusCode, char description[])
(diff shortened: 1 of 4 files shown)


ehpor:
Tested and working on hardware. Each call to `get_power()` takes ~60ms, and is pretty consistent. This is however too slow to call for each exposure separately. The same is true for other devices. This requires more thought; possibly updating the value in a background thread? I'm gonna leave that for a future PR.

jamienoss (approved):
👍  Thanks for adding this.

jamienoss:
I've opened [CATKIT-94](https://jira.stsci.edu/browse/CATKIT-94) to address someone in the future commoning up the utility code common to both this TSP01 code.

PR #228, opened 2021-04-28
HICAT-1083: Take exposures with a generator function

ehpor:
My original plan was to add a `as_generator=False` keyword argument to the `just_take_exposures()` function of the `ZwoCamera`. However, if you have a yield statement in a Python function, even if that yield statement is never executed, the function is still converted to a generator function. Therefore, return statements behave differently, and the original function cannot be used anymore. There are two options left:

* Create a separate generator function that actually takes the exposures, and call that function from `just_take_exposures()`, which then just wraps the images into a list and returns those. This has little impact on existing code, but adds yet another function for taking exposures, which is what we spent a lot of time on to avoid.
* Second option is to convert `just_take_exposures()` into a generator function, and deal with downstream effects. Not a lot of our code is actually calling `just_take_exposures()` directly, so this is by far not as bad as it sounds. This enforces the distinction between `take_exposures()` and `just_take_exposures()` where the latter is meant as a lower-level method of the two.

~I went with the second option for now, as the end result is cleaner.~ I actually went for the first option; see reasons below.

- [x] Hardware tests

### DIFF
--- a/catkit/hardware/zwo/ZwoCamera.py
+++ b/catkit/hardware/zwo/ZwoCamera.py
@@ -225,18 +225,14 @@ class ZwoCamera(Camera):
                        subarray_x=None, subarray_y=None, width=None, height=None, gain=None, full_image=None,
                        bins=None):
         """ Wrapper to take exposures and also save them if `file_mode` is used. """
-        images = []
-        meta = None
 
-        for img, meta in self.just_take_exposures(exposure_time=exposure_time,
-                                                  num_exposures=num_exposures,
-                                                  extra_metadata=extra_metadata,
-                                                  full_image=full_image, subarray_x=subarray_x, subarray_y=subarray_y,
-                                                  width=width, height=height,
-                                                  gain=gain,
-                                                  bins=bins):
-            images.append(img)
-            meta = meta
+        images, meta = self.just_take_exposures(exposure_time=exposure_time,
+                                                num_exposures=num_exposures,
+                                                extra_metadata=extra_metadata,
+                                                full_image=full_image, subarray_x=subarray_x, subarray_y=subarray_y,
+                                                width=width, height=height,
+                                                gain=gain,
+                                                bins=bins)
 
         if file_mode:
             catkit.util.save_images(images, meta, path=path, base_filename=filename, raw_skip=raw_skip)
@@ -247,12 +243,12 @@ class ZwoCamera(Camera):
         else:
             return images
 
-    def just_take_exposures(self, exposure_time, num_exposures,
-                            extra_metadata=None,
-                            subarray_x=None, subarray_y=None, width=None, height=None, gain=None, full_image=None,
-                            bins=None, use_video_capture_mode=True):
+    def stream_exposures(self, exposure_time, num_exposures,
+                         extra_metadata=None,
+                         subarray_x=None, subarray_y=None, width=None, height=None, gain=None, full_image=None,
+                         bins=None, use_video_capture_mode=True):
         """
-        Low level method to take exposures using a Zwo camera. By default keeps image data in.
+        Take exposures and return them using a generator.
 
         :param exposure_time: Pint quantity for exposure time, otherwise in microseconds.
         :param num_exposures: Number of exposures.
@@ -265,7 +261,7 @@ class ZwoCamera(Camera):
         :param full_image: Boolean for whether to take a full image.
         :param bins: Integer value for number of bins.
         :param use_video_capture_mode: Boolean for whether to use video capture or snapshot mode. Default is True.
-        :return: Two parameters: Image list (numpy data or paths), Metadata list of MetaDataEntry objects.
+        :yield: Two parameters: Image (numpy data), Metadata list of MetaDataEntry objects.
         """
 
         # Convert exposure time to contain units if not already a Pint quantity.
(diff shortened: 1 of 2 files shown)


mperrin:
Hi Emiel,  can you explain or clarify how this approach reduces the overheads?  I must be missing something but I would expect that the same hardware overhead and pipeline processing time is present, regardless of the distinction between yield vs return? Is the benefit that the setup overhead can happen once early on and then many images be returned in a series over time?

ehpor:
@mperrin Yes, it removes setup overhead. Also, and this is the main reason, it allows the computer to do processing on that image, calculate a control signal, and put that control signal on the DM *while* the next exposure is taken (if video mode is turned on).

Pseudocode for high-speed control:
```python
# Before this PR:
for i in range(num_iter):
    img, meta = cam.just_take_exposures(num_exposures=1, ...) # Includes setup and overhead for each iteration
    control_signal = calculate_control(img)
    dm.actuators = control_signal

# After this PR:
for img, meta in cam.just_take_exposures(num_exposures=num_iter, ...): # Setup and overhead is done once
    control_signal = calculate_control(img)
    dm.actuators = control_signal
```

Pseudocode for (future) pipeline:
```python
# Before this PR
images, meta = cam.just_take_exposures(...) # Lots of waiting around for exposures to finish
realigned_images = [realign_image_to_first_image(img) for img in images] # All processing is done after exposures are finished

# After this PR
for img, meta in cam.just_take_exposures(...):
    realigned_images.append(realign_image_to_first_image(img) # Most processing is done while taking exposures
```

jamienoss:
@ehpor Do you have an ETA for this or should I go ahead and make a new rel such that you can use the power meter code?

ehpor:
@jamienoss I'm trying to get this done within the next few days. I'm going back on my decision to go for option 2, and will go for option 1 instead. The power meter code is not critical, as that needs more code in hicat to be useful and it needs to fiber splitter to get there, so no rush on a new release from my side.

jamienoss:
>I'm going back on my decision to go for option 2, and will go for option 1 instead.

How comes?

ehpor:
Option 1 has less impact on existing systems. Furthermore, I didn't realize that `just_take_exposures()` is not part of the "official" camera interface (and its use in existing code should therefore be discouraged). So instead of modifying existing code, we should probably aim to just use a new name (like `stream_exposures()`, I'm open to suggestions here) and discontinue using `just_take_exposures()` in the future (maybe even deprecating it now).

jamienoss:
> "official" camera interface

Well... that's neither here nor there as we can just add it to the interface, and probably should have been (but more aptly named).

``just_take_exposures()`` was only added to separate out the file IO.

>we should probably aim to just use a new name (like stream_exposures()

``stream_exposures()`` is a good idea. 👍

ehpor:
Starting CI as a check. Ran fine on hicat-package develop on my machine in simulation. Will do a hardware check now.

ehpor:
Works on hardware, but fails on CI. I'll investigate.

ehpor:
It just ran fine on my machine. I'll restart CI to rule out a statistical fluke (simulated photon noise).

ehpor:
Test failure is caused by the merge of #216. The changing of the sign of segment piston causes a slight drop in total flux, which makes the test fail. Tests pass without #216 merged. Current CI fails as well on develop (tested with #230). @jamienoss What should we do with this? Easiest would be to add a fudge factor to the flux comparison to make everything work again.

Images for this test (left without #216, right with #216):
image

These images look identical to catkit develop before and after #216, as expected.

jamienoss:
@ehpor CI tests for #216 all passed, so it can't _just_ be that PR but something else. I'm not sure that #216 was upto date at the time that I merged it though...

jamienoss:
> These images look identical to catkit develop before and after #216, as expected.

The images are rotated.

ehpor:
> > These images look identical to catkit develop before and after #216, as expected.
> 
> The images are rotated.

These two images are of course not identical to each other, as the piston signs are different (which looks like rotation). These two are however identical to the two images on the current develop (so without this PR). I didn't show those images.

jamienoss:
>What should we do with this? Easiest would be to add a fudge factor to the flux comparison to make everything work again.

We would just change the ``rtol``, however, I think we should really get a handle on whether this change is really as expected or whether it highlights some other issue that has been hiding, and why that PR's tests passed.

However, this is in relation to #216 and not this PR. I'm gonna revert #216 and reopen it - pending CI outcome of #230.

jamienoss:
~@ehpor this branch is out of date and doesn't even include #216.~ Scrap that, too much contention this morning.

ehpor:
> @ehpor this branch is out of date and doesn't even include #216.

That doesn't matter. CI is done on a temporary branch that is this branch merged into develop (refs/remotes/pull/228/merge). This branch is automatically created specifically for CI. The CI isn't done on an unmerged branch.

jamienoss:
#230 downstream tests are passing (https://github.com/spacetelescope/catkit/pull/230/checks?check_run_id=2476179124) so I'm reverting #216.

mperrin:
The change in #216 will certainly change results since it's a different piston pattern. There may be some confusion on #216 - we do want that sign flip (eventually) but I had thought there was code to check the poppy version?  Given the difference in conventions used in poppy 0.9 and 1.0+ that should be checked. Right now I believe (I thought...) catkit and hicat are pinned to poppy 0.9.2, yes?  I had not realized/noticed that the sign flip in 216 wasn't checking against a version number - my apologies for missing that.

jamienoss:
@mperrin there is a version comparison https://github.com/spacetelescope/catkit/pull/216/files#diff-829864cdcbff0a89b9e7a75b45ba7cae2f266aae94b56bcbd9275be24751b5faR55

jamienoss:
@mperrin Let's pause on the convo for a mo, until we open the new PR and we can doc this all there rather than here.

jamienoss:
@ehpor #216 has been reverted so please rebase and restart the CI.

jamienoss:
@ehpor Is this ready to review?

ehpor:
@jamienoss Yes.

jamienoss (approved):
👍

PR #232, opened 2021-04-30
add reversed sign based on poppy version in piston too

jamienoss:
Reopened from #216 (reverted in #231) due to subsequent CI failures in #228 and #230.

See https://github.com/spacetelescope/catkit/runs/2470296681

### DIFF
--- a/catkit/emulators/iris_ao_controller.py
+++ b/catkit/emulators/iris_ao_controller.py
@@ -54,7 +54,7 @@ class PoppyIrisAODM(poppy.dms.HexSegmentedDeformableMirror):
             # Furthermore, depending on poppy version we may need to correct a sign inconsistency
             sign = -1 if Version(poppy.__version__) < Version('1.0') else 1
 
-            piston = values[0] * u.um
+            piston = sign * values[0] * u.um
             tip = sign * values[2] * u.mrad
             tilt = sign * values[1] * u.mrad
             self.set_actuator(seg-1, piston, tip, tilt)    # offset by -1 for 0-based vs 1-based segment indices; see PR #147
--- a/catkit/hardware/iris_ao/segmented_dm_command.py
+++ b/catkit/hardware/iris_ao/segmented_dm_command.py
@@ -494,6 +494,9 @@ def round_ptt_list(ptt_list, decimals=3):
 
 
 def convert_ptt_units(ptt_list, tip_factor, tilt_factor, starting_units, ending_units):
+    # TODO: this should follow the logic from here: https://github.com/spacetelescope/catkit/blob/8c1e22e8f96d51f4453bba27c4710d27da5d37c5/catkit/emulators/iris_ao_controller.py#L50
+    # This is used in display, where it's not super important, and it's used in the global Zernike commands, so those
+    # are not super reliable at this point.
     """
     Convert the PTT list to or from Poppy units and the segmented DM units.
 


jamienoss:
Porting over comments from #228 

@mperrin  (https://github.com/spacetelescope/catkit/pull/228#issuecomment-830127449)
> The change in #216 will certainly change results since it's a different piston pattern. There may be some confusion on #216 - we do want that sign flip (eventually) but I had thought there was code to check the poppy version? Given the difference in conventions used in poppy 0.9 and 1.0+ that should be checked. Right now I believe (I thought...) catkit and hicat are pinned to poppy 0.9.2, yes? I had not realized/noticed that the sign flip in 216 wasn't checking against a version number - my apologies for missing that.

@jamienoss  (https://github.com/spacetelescope/catkit/pull/228#issuecomment-830132790)
>@mperrin there is a version comparison https://github.com/spacetelescope/catkit/pull/216/files#diff-829864cdcbff0a89b9e7a75b45ba7cae2f266aae94b56bcbd9275be24751b5faR55

jamienoss:
Looking at the passing CI logs for #216 (https://github.com/spacetelescope/catkit/runs/2098399742#step:5:677) that Pr was tested with a much older hicat from March 11. It was tested against  664ed37767acc4cb3a453f68de714fab00f30163 which is the merge of #543.

jamienoss:
What's the plan for this?
Who wants to work on it?

ivalaginja:
We need this, @ehpor wanted to look into his dOTF data to confirm that we need it, although both @mperrin and I are very sure we do.
I would take this on, but the failing HiCAT test is about the measured PSF flux not being what is expected from the config, and I don't know how to go about this. Maybe @mperrin could have some pointers that I can use to take care of this? This sign change should be self-consistent in terms of when the inverse flatmap is applied in order to simulate the "relaxed" IrisAO surface together with applying the actual flatmap to flatten it when used. I don't quite understand why that would change the measured flux.

ehpor:
The main issue is that the reversal of sign has the effect of inverting the relaxed state of the Iris. Therefore, the PSF for the relaxed DM will be different, leading to a tiny part of the PSF falling outside of the detector. The test is only failing by a tiny margin (98.8% of total flux falls inside the detector, and the test checks for 100% +/- 1%). If this change is what we want and the change in the relaxed IrisAO surface is expected, then the test simply needs to check for a slightly larger margin (say 1.5%), or we can instead perform the flux test on a flattened Iris. If this change in the PSF is not expected, it signifies something else/bigger going on, see last paragraph.

The curious case of the CI tests passing on #216, I don't really feel like doing archeology, and personally I'd feel okay just fixing the CI tests now, rather than fully knowing why they passed in the past, and fail in the present.

@mperrin @ivalaginja There is a version check. This PR introduces a piston sign flip for Poppy <1.0; for Poppy >=1.0 the sign is not flipped. If Poppy changes the piston sign in 1.0, and this PR is meant to fix that sign flip for when we switch catkit over to Poppy 1.0, then I would expect that version check to be the wrong way around, right? Was that the goal of this PR?

ivalaginja:
> [...] If this change is what we want and the change in the relaxed IrisAO surface is expected, then the test simply needs to check for a slightly larger margin (say 1.5%), or we can instead perform the flux test on a flattened Iris. If this change in the PSF is not expected, it signifies something else/bigger going on, see last paragraph.

Yes, I think that's exactly what we want. I am unsure though if we should allow for a larger margin or just fix the expected number in the first place, which is what we usually do when this test is failing. I think there is a notebook that can calculate the actually expected number, but we'd need to make sure that that notebook uses the correct sign in the IrisAO, otherwise we end up again where we are now.

> @mperrin @ivalaginja There is a version check. This PR introduces a piston sign flip for Poppy <1.0; for Poppy >=1.0 the sign is not flipped. If Poppy changes the piston sign in 1.0, and this PR is meant to fix that sign flip for when we switch catkit over to Poppy 1.0, then I would expect that version check to be the wrong way around, right? Was that the goal of this PR?

The goal of this PR was to bridge the time until poppy 1.0.0+ is released. Poppy 1.0.0 will have the correct signs, which is why for that version, we won't need the sign flip in catkit anymore. We added this bit here to fix that in catkit in the meantime, and the version check makes sure we don't have to do it again once we're tied to poppy 1.0.0+
So the version check is the correct way around.

ivalaginja:
This is the notebook I mean:
https://github.com/spacetelescope/hicat-package/blob/develop/hicat/simulators/Calculate%20sim%20exit_vs_entrance%20pupil%20flux%20ratio%20for%20config.ini.ipynb

As said above, we'd just have to make sure the sign is actually correct when using the notebook if we expect it to give us a more accurate number that we intend to use in the CI.

mperrin:
OK I'll ask the really basic question that will show how out of it I am now. Do we have unambiguous documentation somewhere of the sign of the move in the real hardware for a positive piston command? Does a positive command move the actuator forward into the beam (reducing path length, positive OPD) or backwards towards the DM substrate and drive electronics (increasing path length, negative OPD).

ivalaginja:
@mperrin I don't know about forward and backward in the beam, but I do know that it is currently not doing the same in sim and on hardware when sending the exact same command. Here is some data - the data *includes* the sign fix we added in this PR:

The top row are hardware images, bottom are simulated images. The left column contains images with the IrisAO flattened, without the FPM in the beam, and the PSF center is marked by the red cross. The right column contains image in which the IrisAO segments have been pistoned in such a way that we are approximating a global tilt with 1 lambda ptv WFE. The PSF center of this PSF tilted with local pistons is indicated with the blue cross. The red plus and blue cross in a pair of images are exactly the same values, so that we can compare them in each image individually.

We can see that the PSF is tilted in the same direction on the detector when comparing the hardware and simulated data, *when the sign fix is added*. I have checked that the simulated PSF indeed moves in the other direction when I perform the same experiment without the added sign flip.

Screenshot 2021-06-03 at 21 22 16

Is this enough to confirm that we need this sign flip? Another thing I noticed is that the PSF does *not* move by the same amount between hardware and simulations when I used exactly the same numbers as input, which is indicated with the numbers circled in red. These numbers are the difference in PSF centers between the reference PSF and the tilted PSF along the x-axis. This suggests that we are missing a factor of two in the simulations when commanding the IrisAO segments. This makes sense to me: we know that the IrisAO controller keeps its commands in units of surface, and we know that poppy keeps everything consistently in WFE, which we haven't been taking into account so far. This would be fixed if we include `include_factor_of_two=True` [where the simulated IrisAO gets created](https://github.com/spacetelescope/hicat-package/blob/d475f4ce7ac27664625b5c893d9bd79cce091a93/hicat/simulators/hicat_sim_poppy.py#L139), but I didn't run any tests yet to confirm this for tip/tilt as well (although it would be very surprising if it is not the case for them as well...)

mperrin (approved):
Based on Iva's test, described in the comment and figures above, and a conversation with Iva just now, I am now entirely convinced this sign flip is needed (as expected!).   So I approve this PR.

If this ends up again causing some CI issue, we should debug and fix that, but this PR at this point seems clearly correct. 

There will need to be a separate PR for the factor_of_two as well.

kjbrooks (commented):
A comment from the peanut gallery: I completely agree with Iva's assessment of the factor of two, which I have seen in tip and tilt as well in GS stuff (for the next PR, of course).

ivalaginja:
@kjbrooks oh this is great!! There's our data that confirms this then. I am about to open a PR on hicat to fix it.

ivalaginja:
The catkit tests pass locally, the hicat tests do not. Looking into it now.

ivalaginja:
> This is the notebook I mean:
> https://github.com/spacetelescope/hicat-package/blob/develop/hicat/simulators/Calculate%20sim%20exit_vs_entrance%20pupil%20flux%20ratio%20for%20config.ini.ipynb
> 
> As said above, we'd just have to make sure the sign is actually correct when using the notebook if we expect it to give us a more accurate number that we intend to use in the CI.

This comment was wrong by the way, I had mixed up different configfile parameters.

ivalaginja:
Also, the test that is failing is still using `run_hicat_imaging()` 😭

ivalaginja:
I would like to ask for help on debugging the failing test please @mperrin @jamienoss. We know from @ehpor above that the problem is a slightly different flux, but we're not sure why. Emiel is right that this comes most likely from a small change in the PSF, but I don't think such a change is actually expected form the changes in this PR. I say this because the "relaxed" IrisAO in sim is just an inverted flatmap, so once we apply the flatmap it should be perfectly zero. Our changes simply change what that flatmap looks like, so I wouldn't expect a change in the resulting PSF (from the zeroed IrisAO).

I also can't run this test by itself since it accesses Devices and those need to be run from an experiment. Does anybody have ideas on how to deal with this?

ivalaginja:
Also, I think I want to replace run_hicat_iamging with 
```py
psf_data, psf_meta = Device.IMAGING_CAMERA.just_take_exposures(exptime, 1)
```
but I don't know how to access specific entries of the metadata list that gets returned by just_take_exposures() - the test wants to check whether the header/metadata includes a "PATH" entry.

jamienoss:
@ivalaginja RE using ``run_hicat_iamging()`` vs ``just_take_exposures()`` - we could add another test that does that, and yep, you're prob correct that it  _should_, however, if this test really has exposed a bug (or at least an unwanted sensitivity), I think it would be a good idea to keep it. Until ``run_hicat_iamging()`` gets deprecated, assuming it shouldn't already be 🤷 

BTW ``run_hicat_iamging()`` doesn't flatten any of the DMs.

For example the following test passes:

```python
def test_testbed_interface_flux_just_take_exposure(tmpdir, disable_sim_camera_noise_model):
    """ Test that the 'testbed' interface output PSF in direct mode
    has a flux matching our expectations
    """

    exptime_microsec = 1000
    exptime = quantity(exptime_microsec, units.microsecond)

######### NEW
    from hicat.hardware.testbed import Device
    from catkit.catkit_types import FpmPosition, LyotStopPosition, FlipMountPosition
    from hicat.experiments.modules import iris_ao

    Device.FPM_MOUNT.move(FpmPosition.OUT_OF_BEAM)  # Comparing direct photometry.
    Device.IRIS_DM.apply_shape(iris_ao.flat_command())
    psf_data, psf_header = Device.IMAGING_CAMERA.just_take_exposures(exptime, 1)
############

    psf_image = psf_data[0]

    counts_per_microsec = CONFIG_INI.getfloat(
        'photometry', 'total_direct_photometry_cts_per_microsec', fallback=90000)

    flux_calc, flux_expected = psf_image.sum(), counts_per_microsec * exptime_microsec
    assert np.allclose(flux_calc, flux_expected, rtol=0.01), f"PSF flux not as expected. " \
                                                             f"Sim result is {flux_calc} but expected {flux_expected} based on config settings."
```

``test_testbed_interface_flux`` also passes when you flatten the Iris before calling ``run_hicat_iamging()``.

ivalaginja:
Ok... my next question would then be, how was this test passing before, if it wasn't flatting the DMs? I am confused as to we shouldn't have caught this earlier, literally every single time we were running these tests.
Thanks a lot for testing this @jamienoss. I think the simplest way forward then is to flatten the IrisAO in the test we currently have. Still confused at the fact that this is not failing outside of this PR though...

ehpor:
@ivalaginja 
> Still confused at the fact that this is not failing outside of this PR though...

The test was loose enough to just about not fail with the relaxed Iris. The inversion of the segment pistons just pushed it over the edge.

ivalaginja:
That makes sense.
I added a line to flatten the IrisAO now (on the HiCAT side), tests are passing locally, PR for that open under https://github.com/spacetelescope/hicat-package/pull/592

ivalaginja:
https://github.com/spacetelescope/hicat-package/pull/592 is merged in HiCAT so lets retrigger the tests here.

kjbrooks (approved):

PR #236, opened 2021-05-12
CATKIT-96: Reduce file IO overhead for sending commands to BMC

ehpor:
This PR caches the flat maps and dm mask which were read every time a DM command was sent to the Boston. This reduces the timing:

Before this PR:
My machine on simulator: 6.4-7.0ms.
Hicat2 on hardware: 4.7-4.8 ms.

After this PR:
My machine on simulator: 1.0ms.
Hicat2 on hardware: 0.71-0.81 ms.

- [X] Tested strokemin and performance on simulator.
- [x] Tested performance on hardware.
- [x] Full strokemin on hardware? (is this necessary?)

### DIFF
--- a/catkit/hardware/boston/DmCommand.py
+++ b/catkit/hardware/boston/DmCommand.py
@@ -12,6 +12,8 @@ import catkit.util
 m_per_volt_map1 = None  # for caching the conversion factor, to avoid reading from disk each time
 m_per_volt_map2 = None  # for caching the conversion factor, to avoid reading from disk each time
 
+flat_map_dm1 = None  # for caching the conversion factor, to avoid reading from disk each time
+flat_map_dm2 = None  # for caching the conversion factor, to avoid reading from disk each time
 
 class DmCommand(object):
     def __init__(self, data, dm_num, flat_map=False, bias=False, as_voltage_percentage=False,
@@ -107,14 +109,7 @@ class DmCommand(object):
 
             # OR apply Flat Map (which is itself biased).
             elif self.flat_map:
-                if self.dm_num == 1:
-                    flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm1")
-                    flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
-                    dm_command += flat_map_volts[0].data
-                else:
-                    flat_map_file_name = CONFIG_INI.get("boston_kilo952", "flat_map_dm2")
-                    flat_map_volts = fits.open(os.path.join(self.calibration_data_path, flat_map_file_name))
-                    dm_command += flat_map_volts[0].data
+                dm_command += get_flat_map_volts(self.dm_num)
 
             # Convert between 0-1.
             dm_command /= self.max_volts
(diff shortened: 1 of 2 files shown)


catkit/hardware/boston/DmCommand.py
```diff
@@ -191,14 +186,18 @@ def get_flat_map_volts(dm_num):
...
+
+    global flat_map_dm1, flat_map_dm2
+
+    if flat_map_dm1 is None:
+        fname = CONFIG_INI.get("boston_kilo952", "flat_map_dm1")
+        flat_map_dm1 = fits.getdata(os.path.join(calibration_data_path, fname))
+
+    if flat_map_dm2 is None:
+        fname = CONFIG_INI.get("boston_kilo952", "flat_map_dm2")
+        flat_map_dm2 = fits.getdata(os.path.join(calibration_data_path, fname))
```
jamienoss: I would still load them conditionally to account for single DM usage. There's no strong reason to do so, only the weak example case of us messing around with a single DM and not even having the second file on disk.

catkit/util.py
```diff
@@ -25,14 +25,19 @@ def find_package_location(package='catkit'):
     return importlib.util.find_spec(package).submodule_search_locations[0]
 
 
-def find_repo_location(package='cakit'):
+def find_repo_location(package='catkit'):
```
jamienoss, line 28: Good catch, thanks. 👍

catkit/util.py
```diff
@@ -25,14 +25,19 @@ def find_package_location(package='catkit'):
...
     return os.path.abspath(os.path.join(find_package_location(package), os.pardir))
 
 
 def get_dm_mask():
-    mask_path = os.path.join(find_package_location("catkit"), "hardware", "boston", "kiloCdm_2Dmask.fits")
-    mask = fits.open(mask_path)[0].data
-    return mask
+    if get_dm_mask.mask is None:
```
jamienoss: Let's do the following just to keep code contained to the func def, please.

```python
if getattr(get_dm_mask, "mask", None) is None:
  ...
```

catkit/util.py
```diff
@@ -25,14 +25,19 @@ def find_package_location(package='catkit'):
...
+        mask_path = os.path.join(find_package_location("catkit"), "hardware", "boston", "kiloCdm_2Dmask.fits")
+        get_dm_mask.mask = fits.getdata(mask_path)
+
+    return get_dm_mask.mask
+
+
+get_dm_mask.mask = None
```
jamienoss: See comment above.

catkit/hardware/boston/DmCommand.py
```diff
@@ -262,26 +261,18 @@ def convert_m_to_volts(data, dm_num, meter_to_volt_map=None):
...
-    index952 = np.flatnonzero(mask)
+    mask = catkit.util.get_dm_mask().astype('bool')
+    number_of_actuators = np.count_nonzero(mask)
 
-    number_of_actuators_per_dimension = CONFIG_INI.getint('boston_kilo952', 'dm_length_actuators')
-    number_of_actuators = CONFIG_INI.getint("boston_kilo952", "number_of_actuators")
-    image = np.zeros((number_of_actuators_per_dimension, number_of_actuators_per_dimension))
-    image[np.unravel_index(index952, image.shape)] = dm_command[:number_of_actuators]
+    image = np.zeros(mask.shape)
+    image[mask] = dm_command[:number_of_actuators]
```
jamienoss, line 287: 👍  Good reduction! It took me way too long to even figure out what the old code was doing.

jamienoss (changes requested):
Thanks for doing this, just some minor changes, please, thanks.

jamienoss:
@ehpor RE
> Full strokemin on hardware? (is this necessary?)

Yes, please, thanks.

catkit/hardware/boston/DmCommand.py
ehpor: I'll do the same for the `get_m_per_volt_map()` function just below this one then.

ehpor:
Tested on hardware: `2021-05-13T15-12-07_broadband_stroke_minimization`.

catkit/hardware/boston/DmCommand.py
```diff
@@ -182,22 +182,33 @@ def load_dm_command(path, dm_num=1, flat_map=False, bias=False, as_volts=False):
...
-        flat_map_dm2 = fits.getdata(os.path.join(calibration_data_path, fname))
+    if dm_num == 2:
+        if flat_map_dm2 is None:
+            fname = CONFIG_INI.get("boston_kilo952", "flat_map_dm2")
+            flat_map_dm2 = fits.getdata(os.path.join(calibration_data_path, fname))
 
-    return flat_map_dm1 if dm_num == 1 else flat_map_dm2
+        return flat_map_dm2
+
+    raise ValueError('dm_num should be either 1 or 2.')
```
jamienoss, line 211: Not gonna ask you to redo this, but this sort of logic pattern is neater and more resilient to future changes when self-contained as an ``else`` to an ``elif``, i.e., 
```python
if dm_num == 1:
  ...
elif dm_num == 2:
  ...
else:
   raise
```

jamienoss (approved):

catkit/hardware/boston/DmCommand.py
ehpor: I considered doing a dictionary approach too. That would have been even cleaner. PEP 634 is gonna be so good for this type of stuff too. I've been waiting for that for a long time.

PR #238, opened 2021-05-26
HICAT-100 Auto port feature for shared memory  server

jamienoss:
Upstream PR for https://github.com/spacetelescope/hicat-package/pull/589

Feature suggested by @ehpor.

I'm not sure if I like this as is, gonna leave it here as a draft for now till I get back to it for the multiple device servers.


jamienoss:
~Might be better to not introduce another bool but have a neg port value := auto, or just None.~ This is already a feature of the base manager, we can just pass in 0 for the port and it'll dish one out.

And we should just do this as default for sim mode (but only those form ``CONFIG_INI`` so it doesn't mess up the tests).

PR #240, opened 2021-09-28
Adding channels to the BMC DM.

ehpor:
This adds channels to the BMC DM.

Channels act as individual contributions to the DM shape. One channel can be updated at a time, without disturbing other contributions to the DM. This allows multiple loops to apply corrections to the DM, without knowing about the state of other loops. A channel requires a name (as string). Using `None` as the channel (default), rewrites the whole DM shape and doesn't take into account any set channel contributions. This way, everything is backwards compatible.

This PR also adds a lock / mutex to the `apply_shape_to_both()` function. This is necessary for multiple loops running in the same process and sending commands to the DM at potentially the same time. Access to the internal memory for channels is serialized.

I also made logging optional (opt-out; default is still to emit log messages). Having a loop send commands to the BMC at >100Hz makes the log files quite unreadable without.

### DIFF
--- a/catkit/hardware/boston/BostonDmController.py
+++ b/catkit/hardware/boston/BostonDmController.py
@@ -114,7 +114,7 @@ class BostonDmController(DeformableMirrorController):
                             as_volts=False,
                             sin_specification=None,
                             output_path=None,
-                            channel=None
+                            channel=None,
                             do_logging=True):
         """ Combines both commands and sends to the controller to produce a shape on each DM.
         :param dm<1|2>_shape: catkit.hardware.boston.DmCommand.DmCommand or numpy array of the following shapes: 34x34, 1x952,
--- a/catkit/hardware/boston/BostonDmController.py
+++ b/catkit/hardware/boston/BostonDmController.py
@@ -35,7 +35,6 @@ class BostonDmController(DeformableMirrorController):
         self.dm2_command_object = None
 
         self.channels = {}
-        self.lock = threading.Lock()
 
     def initialize(self, serial_number, command_length, dac_bit_width):
         """ Initialize dm manufacturer specific object - this does not, nor should it, open a connection."""
@@ -45,6 +44,8 @@ class BostonDmController(DeformableMirrorController):
         self.serial_num = serial_number
         self.command_length = command_length
         self.dac_bit_width = dac_bit_width
+
+        self.lock = threading.Lock()
 
     def send_data(self, data):
 
@@ -117,6 +118,19 @@ class BostonDmController(DeformableMirrorController):
                             channel=None,
                             do_logging=True):
         """ Combines both commands and sends to the controller to produce a shape on each DM.
+
+        The concept of channels is optional. If no channel is supplied, the DM shapes are applied directly
+        onto the DM. However, if channels are used, each channel acts as an independent contribution to the
+        total shape that is on the DM. Each contribution will be updated (= replaced) by calling apply_shape_to_both()
+        with the name of that channel. In this way, the current contribution from each channel can be read out
+        using the BOSTON_DM.channels[channel_name] attribute.
+
+        While individual contributions can be added as delta contributions to a running total, this approach was
+        not taken for code clarity. This comes at the cost of a few microseconds of runtime for each sent DM command.
+
+        Note: if channels are used, the dm shapes are required to be numpy arrays. In this case, DmCommand objects are
+        not allowed. A TypeError will be thrown is this is the case.
+
         :param dm<1|2>_shape: catkit.hardware.boston.DmCommand.DmCommand or numpy array of the following shapes: 34x34, 1x952,
                          1x2048, 1x4096. Interpreted by default as the desired DM surface height in units of meters, but
                          see parameters as_volts and as_voltage_percentage. When using channels, this should be a numpy array,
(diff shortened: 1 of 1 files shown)


catkit/hardware/boston/BostonDmController.py
```diff
@@ -109,64 +113,96 @@ def apply_shape_to_both(self, dm1_shape, dm2_shape,
...
+            if channel is None:
+                self.log.info("Applying shape to both DMs")
+            else:
+                self.log.info(f'Applying shape to both DMs in channel {channel}.')
+
+        with self.lock:
+            if channel is not None:
+                if isinstance(dm1_shape, DmCommand) or isinstance(dm2_shape, DmCommand):
+                    # DmCommand objects cannot be added together, yet.
+                    raise ValueError('DM shapes cannot be DmCommands when using channels.')
```
jamienoss: [revision] This is actually a ``TypeError`` (or ``NotImplementedError``).

catkit/hardware/boston/BostonDmController.py
```diff
@@ -33,6 +34,9 @@ def _clear_state(self):
         self.dm1_command_object = None
         self.dm2_command_object = None
 
+        self.channels = {}
+        self.lock = threading.Lock()
```
jamienoss: [revision] Hoist this out to ``Initialize()``, we don't want a scenario where there are multiple locks.

catkit/hardware/boston/BostonDmController.py
```diff
@@ -109,64 +113,96 @@ def apply_shape_to_both(self, dm1_shape, dm2_shape,
...
-            self.dm2_command = dm2_command
-            self.dm1_command_object = dm1_shape
-            self.dm2_command_object = dm2_shape
+        if do_logging:
+            if channel is None:
+                self.log.info("Applying shape to both DMs")
+            else:
+                self.log.info(f'Applying shape to both DMs in channel {channel}.')
+
+        with self.lock:
```
jamienoss: [revision] Might as well add the logging to the crit-sec since we pretty much mutex the entire func call. Would improve debugging (maybe).

catkit/hardware/boston/BostonDmController.py
```diff
@@ -109,64 +113,96 @@ def apply_shape_to_both(self, dm1_shape, dm2_shape,
...
+
+                self.channels[channel] = (dm1_shape, dm2_shape)
+
+                # Add contributions for each channel, and use that as the dm command.
+                dm1_shape = 0
+                dm2_shape = 0
+
+                for dm1, dm2 in self.channels.values():
+                    dm1_shape += dm1
+                    dm2_shape += dm2
```
jamienoss, line 181: ~[design revision?] Since the func sig allows only one ch per call, and everything's commutative,  doesn't it make more sense to add a channel's contribution to a running total (i.e., the actual command/image) rather than stashing each ch individually and then having to re-sum them every time (even though only one has changed)?~

Now I remember, what I've described is the "delta API" that I've mentioned before but you wanted the chs readable and stashed in the device obj rather than by the calling thread/process.

[revision] Can we add a comment outlining the design, please, at least so I don't chase my tail again.

[revision] When ``channel is None``, do we not need to clear the channels? Because, if the reason for keeping them separate is to be able to read them and then act upon them, well... they'll be stale, right? I'm pretty sure we should be resetting this.

catkit/hardware/boston/BostonDmController.py
```diff
@@ -109,64 +113,96 @@ def apply_shape_to_both(self, dm1_shape, dm2_shape,
...
+            if channel is not None:
+                if isinstance(dm1_shape, DmCommand) or isinstance(dm2_shape, DmCommand):
+                    # DmCommand objects cannot be added together, yet.
+                    raise ValueError('DM shapes cannot be DmCommands when using channels.')
+
+                # Check if dm{1,2}_shape is 2D, then convert to 1D.
+                if dm1_shape.ndim == 2:
+                    dm1_shape = convert_dm_image_to_command(dm1_shape)
+                if dm2_shape.ndim == 2:
+                    dm2_shape = convert_dm_image_to_command(dm2_shape)
```
jamienoss, line 171: This is the part that threw me at first as ``convert_dm_image_to_command()`` seemed like the functionality _of_ ``DmCommand``. It seemed redundant to do this before creating a ``DmCommand``.

What's the rationale for flattening these here, is it just in case another shape already comes in flat so we're standardizing?

jamienoss (changes requested):
Thanks for adding this. There are a few revisions and some general questions but they can be punted to another pr/issue.

catkit/hardware/boston/BostonDmController.py
ehpor: Fixed.

catkit/hardware/boston/BostonDmController.py
ehpor: Fixed.

catkit/hardware/boston/BostonDmController.py
ehpor: Fixed.

catkit/hardware/boston/BostonDmController.py
ehpor: Comment added in the docstring of the function.

I'm now clearing the channels if `channel is None`. I'm also emitting a warning to the log, because this is extremely likely due to misuse of the channels API.

catkit/hardware/boston/BostonDmController.py
ehpor: Indeed this is for standardization of the stored numpy arrays for the DM shapes. I've added a comment to that effect in the code.

jamienoss (approved):

PR #242, opened 2021-11-30
Add FLIR camera device.

ehpor:
This PR does _not_ contain an emulator for a FLIR camera. It can therefore not be used with the HiCAT simulator. This will be done in a separate PR.

- [x] Add docstrings.
- [ ] Hardware test (worked before the rebase, and the rebase was clean as well, so just a quick check).

### DIFF
--- a/catkit/hardware/flir/flir_camera.py
+++ b/catkit/hardware/flir/flir_camera.py
@@ -129,6 +129,11 @@ class FlirCamera(Camera):
         self.cam.AcquisitionMode.SetValue(self.instrument_lib.AcquisitionMode_Continuous)
         #self.cam.StreamBufferHandlingMode.SetValue(self.instrument_lib.StreamBufferHandlingMode_NewestOnly)
 
+        nodemap = self.cam.GetTLStreamNodeMap()
+        node = self.instrument_lib.CEnumerationPtr(nodemap.GetNode('StreamBufferHandlingMode'))
+        newest_only = node.GetEntryByName('NewestOnly')
+        node.SetIntValue(newest_only.GetValue())
+
         # Turn off triggering
         self.cam.TriggerMode.SetValue(self.instrument_lib.TriggerMode_Off)
 
--- a/catkit/hardware/flir/flir_camera.py
+++ b/catkit/hardware/flir/flir_camera.py
@@ -34,6 +34,21 @@ class FlirLibrary(metaclass=LazyLoadLibraryMeta):
 
 
 def _create_property(flir_property_name, read_only=False):
+    '''Helper for creating non-enum PySpin properties.
+
+    Parameters
+    ----------
+    flir_property_name : string
+        The name of the PySpin property. This property must be accessible from
+        the PySpin Camera object.
+    read_only : boolean
+        Whether to make the property read-only or not.
+
+    Returns
+    -------
+    property
+        The created Python property.
+    '''
     def getter(self):
         return getattr(self.cam, flir_property_name).GetValue()
 
@@ -47,6 +62,21 @@ def _create_property(flir_property_name, read_only=False):
 
 
 def _create_enum_property(flir_property_name, enum_name):
+    '''Helper for creating enum PySpin properties.
+
+    Parameters
+    ----------
+    flir_property_name : string
+        The name of the PySpin property. This property must be accessible from
+        the PySpin Camera object.
+    enum_name : dictionary
+        A dictionary describing the mapping from string (key) to PySpin enum value (value).
+
+    Returns
+    -------
+    property
+        The created Python property.
+    '''
     def getter(self):
         value = getattr(self.cam, flir_property_name).GetValue()
 
@@ -72,6 +102,8 @@ class FlirCamera(Camera):
 
     @classmethod
     def _create_system(cls):
+        '''Create the PySpin system instance and keep ref count for tidy release.
+        '''
         if cls._system is None:
             cls._system = cls.instrument_lib.System.GetInstance()
 
@@ -79,6 +111,9 @@ class FlirCamera(Camera):
 
     @classmethod
     def _destroy_system(cls):
+        '''Destroy our reference to the PySpin system instance and release it if there
+        are no references to it anymore.
+        '''
         cls._system_ref_count -= 1
 
         if cls._system_ref_count == 0:
@@ -162,6 +197,7 @@ class FlirCamera(Camera):
 
         self._destroy_system()
 
+    # Exposure time in us.
     exposure_time = _create_property('ExposureTime')
     gain = _create_property('Gain')
 
@@ -170,19 +206,50 @@ class FlirCamera(Camera):
     offset_x = _create_property('OffsetX')
     offset_y = _create_property('OffsetY')
 
+    # Temperature of the internal temperature sensor in degrees Celsius.
     temperature = _create_property('DeviceTemperature', read_only=True)
 
+    # Pixel format (for the transport of data from the camera to the computer)
     pixel_format = _create_enum_property('PixelFormat', 'pixel_format_enum')
+
+    # Bit depth of the ADC (8bit, 12bit, etc...)
     adc_bit_depth = _create_enum_property('AdcBitDepth', 'adc_bit_depth_enum')
 
+    # Allows for modifying the duty cycle of the camera. If disabled, the camera runs
+    # at 100% duty cycle (excluding sensor readout), regardless of the value of
+    # `acquisition_frame_rate`.
     acquisition_frame_rate = _create_property('AcquisitionFrameRate')
     acquisition_frame_rate_enable = _create_property('AcquisitionFrameRateEnable')
 
     @property
     def device_name(self):
+        '''The model of the camera (eg. "BFS-U3-04S2M-C").
+        '''
         return self.cam.TLDevice.DeviceModelName.GetValue()
 
     def stream_exposures(self, exposure_time, num_exposures, extra_metadata=None):
+        '''Stream exposures from the camera.
+
+        Note: contrary to the ZWO camera, all subarray settings must be set on the
+        camera before this generator is called.
+
+        Parameters
+        ----------
+        exposure time : scalar or pint quantity
+            The exposure time for each frame (in us if this is a scalar).
+        num_exposures : int
+            The number of exposures to yield. It is safe to stop this generator before
+            this number of exposures is reached.
+        extra_metadata : list of MetaDataEntry objects
+            The metadata to attach to each output frame.
+
+        Yields
+        ------
+        img : ndarray
+            Each camera frame separately.
+        meta_data : list of MetaDataEntry objects
+            The metadata for each frame.
+        '''
         if not type(exposure_time) in [int, float]:
             exposure_time_us = exposure_time.to(units.microsecond).m
         else:
@@ -191,7 +258,8 @@ class FlirCamera(Camera):
         self.exposure_time = exposure_time_us
 
         meta_data = [MetaDataEntry("Exposure Time", "EXP_TIME", exposure_time_us, "microseconds")]
-        meta_data.append(MetaDataEntry("Camera", "CAMERA", self.device_name, "Camera model, correlates to entry in ini"))
+        meta_data.append(MetaDataEntry("Camera", "CAMERA", self.config_id, "Camera name, correlates to entry in ini"))
+        meta_data.append(MetaDataEntry("CameraModel", "MODEL", self.device_name, "Camera model name"))
         meta_data.append(MetaDataEntry("Gain", "GAIN", self.gain, "Gain for camera"))
 
         if extra_metadata is not None:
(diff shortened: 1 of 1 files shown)


catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+                                  'mono12p': self.instrument_lib.PixelFormat_Mono12p,
+                                  'mono16': self.instrument_lib.PixelFormat_Mono16}
+
+        self.adc_bit_depth_enum = {'8bit': self.instrument_lib.AdcBitDepth_Bit8,
+                                   '10bit': self.instrument_lib.AdcBitDepth_Bit10,
+                                   '12bit': self.instrument_lib.AdcBitDepth_Bit12,
+                                   '14bit': self.instrument_lib.AdcBitDepth_Bit14}
+
+    def _open(self):
+        serial_number = CONFIG_INI.get(self.config_id, 'serial_number')
```
jamienoss, line 137: [revision] We don't ref ``CONFIG_INI`` from ``catkit`` so ``serial_number`` will need to be passed in.

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def _close(self):
+        try:
+            self.cam.EndAcquisition()
+        except self.instrument_lib.SpinnakerException as e:
+            if e.errorcode == -1002:
+                # Camera was not running. We can safely ignore.
+                pass
+            else:
+                self.log.error(f'PySpin error: {e.errorcode}')
```
jamienoss, line 192: [revision?] Do we not want to raise this?

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+
+        self.cam.DeInit()
+        self.cam = None
+        self.cam_list.Clear()
+
+        self._destroy_system()
```
jamienoss, line 198: [revision] These look like they should be in a ``finally`` block, perhaps even two:
```python
...
  finally:
    self.cam.DeInit()  # <----- Can this raise?
finally:
  self.cam = None
  self.cam_list.Clear()
  self._destroy_system()
```

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
+import importlib
+import logging
+
+import numpy as np
+
+from catkit.interfaces.Camera import Camera
+from catkit.catkit_types import MetaDataEntry, units, quantity
+import catkit.util
+from catkit.config import CONFIG_INI
```
jamienoss, line 9: [revision] Remove this, please, we don't ref ``CONFIG_INI`` from catkit.[revision] We don't ref ``CONFIG_INI`` from ``catkit``.

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+
+        # Set up enum dictionaries
+        self.pixel_format_enum = {'mono8': self.instrument_lib.PixelFormat_Mono8,
+                                  'mono12p': self.instrument_lib.PixelFormat_Mono12p,
+                                  'mono16': self.instrument_lib.PixelFormat_Mono16}
+
+        self.adc_bit_depth_enum = {'8bit': self.instrument_lib.AdcBitDepth_Bit8,
+                                   '10bit': self.instrument_lib.AdcBitDepth_Bit10,
+                                   '12bit': self.instrument_lib.AdcBitDepth_Bit12,
+                                   '14bit': self.instrument_lib.AdcBitDepth_Bit14}
```
jamienoss, line 134: [revision] If we're gonna mangle "enum" semantics into the name, then I'd rather see these as actual enums, please (or remove "enum" from the attr name):
```python
self.pixel_format_enum = Enum("pixel_format_enum", {'MONO8': self.instrument_lib.PixelFormat_Mono8,
  'MONO12_P': self.instrument_lib.PixelFormat_Mono12p,
  'MONO16': self.instrument_lib.PixelFormat_Mono16})
```

The reason is that the access pattern between a ``dict`` and an ``enum`` is quite different.

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+        # Obtain singleton Flir system object
+        self._create_system()
+
+        self.cam_list = self._system.GetCameras()
+        self.cam = self.cam_list.GetBySerial(serial_number)
+
+        try:
+            self.cam.Init()
+        except:
+            raise RuntimeError(f'Error during intialization of {self.config_id}. Is the camera connected?')
```
jamienoss, line 148: [revision] Chain these, please.
```python
...
except Exception as error:
  raise RuntimeError("...") from error
```

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+                                   '14bit': self.instrument_lib.AdcBitDepth_Bit14}
+
+    def _open(self):
+        serial_number = CONFIG_INI.get(self.config_id, 'serial_number')
+
+        # Obtain singleton Flir system object
+        self._create_system()
+
+        self.cam_list = self._system.GetCameras()
+        self.cam = self.cam_list.GetBySerial(serial_number)
```
jamienoss, line 143: Do you want to check if this is ``None`` or have you just deferred that to ``self.cam.Init()`` below?

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+    @classmethod
+    def _destroy_system(cls):
+        '''Destroy our reference to the PySpin system instance and release it if there
+        are no references to it anymore.
+        '''
+        cls._system_ref_count -= 1
+
+        if cls._system_ref_count == 0:
+            cls._system.ReleaseInstance()
+            cls._system = None
```
jamienoss, line 121: [revision] Unless the doc states that ``ReleaseInstance()`` never raises, these should be split out into a ``try-finally``. 
```python
try:
  cls._system.ReleaseInstance()
finally:
  cls._system = None
```

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+    property
+        The created Python property.
+    '''
+    def getter(self):
+        value = getattr(self.cam, flir_property_name).GetValue()
+
+        # Reverse search in enum dictionary.
+        for key, val in getattr(self, enum_name).items():
+            if value == val:
+                return key
```
jamienoss, line 86: So, given your use case on [line 213](https://github.com/spacetelescope/catkit/pull/242/files#diff-5279d10039d460f6f63958482f51515f2429dfd3ee8e362c579253440d0e7e9aR213), if they were actual enums you would just do something like:
```
return getattr(self, enum_name)(value)._name_
```

[revision] This (in conjunction with other comments/revisions) is really starting to look like you want to actually use ``enum.Enum``.

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+        self.width = int(CONFIG_INI.get(self.config_id, 'width'))
+        self.height = int(CONFIG_INI.get(self.config_id, 'height'))
+        self.offset_x = int(CONFIG_INI.get(self.config_id, 'offset_x'))
+        self.offset_y = int(CONFIG_INI.get(self.config_id, 'offset_y'))
+
+        self.pixel_format = CONFIG_INI.get(self.config_id, 'pixel_format')
+        self.adc_bit_depth = CONFIG_INI.get(self.config_id, 'adc_bit_depth')
+
+        self.exposure_time = float(CONFIG_INI.get(self.config_id, 'exposure_time'))
+        self.gain = float(CONFIG_INI.get(self.config_id, 'gain'))
```
jamienoss, line 179: [revision} See early revision regarding ``CONFIG_INI``.

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+        self._destroy_system()
+
+    # Exposure time in us.
+    exposure_time = _create_property('ExposureTime')
+    gain = _create_property('Gain')
+
+    width = _create_property('Width')
+    height = _create_property('Height')
+    offset_x = _create_property('OffsetX')
+    offset_y = _create_property('OffsetY')
```
jamienoss, line 207: [revision] These "flir_property_names" look like strict literals dictated by the FLIR doc, right? If so, they should be grouped as a module enum:
```python
class FLIRAttrs(Enum):
  WIDTH = "Width"
  HEIGHT = "Height"
  ...  # etc
```

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+            Each camera frame separately.
+        meta_data : list of MetaDataEntry objects
+            The metadata for each frame.
+        '''
+        if not type(exposure_time) in [int, float]:
+            exposure_time_us = exposure_time.to(units.microsecond).m
+        else:
+            exposure_time_us = exposure_time
+
+        self.exposure_time = exposure_time_us
```
jamienoss, line 258: [revision] Use ``isinstance`` and not ``type`` for this, please.

[revision] Conditional assignment is preferred here, please, or at least dedupe ``exposure_time_us``.

```python
self.exposure_time = exposure_time.to(units.microsecond).m if isinstance(exposure_time, (int, float)) else exposure_time
```
or
```python
self.exposure_time = exposure_time
if not isinstance(self.exposure_time, (int, float)):
  self.exposure_time = self.exposure_time.to(units.microsecond).m
```

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+        meta_data.append(MetaDataEntry("CameraModel", "MODEL", self.device_name, "Camera model name"))
+        meta_data.append(MetaDataEntry("Gain", "GAIN", self.gain, "Gain for camera"))
+
+        if extra_metadata is not None:
+            if isinstance(extra_metadata, list):
+                meta_data.extend(extra_metadata)
+            else:
+                meta_data.append(extra_metadata)
+
+        # try-finally shuts down the camera if the generator is stopped.
```
jamienoss, line 271: Does ``EndAcquisition()`` really _"shut down"_ the camera?

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+            # Get the pixel format to interpret the raw data.
+            if self.pixel_format == 'mono8':
+                pixel_format = self.instrument_lib.PixelFormat_Mono8
+            else:
+                pixel_format = self.instrument_lib.PixelFormat_Mono16
+
+            while frame_count < num_exposures:
+                try:
+                    # Wait for image with timeout (in ms).
+                    image_result = self.cam.GetNextImage(100)
```
jamienoss, line 285: [revision] Let's hoist the timeout to an instance param, please. Or perhaps a class attr if there's a reason to.

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+                except self.instrument_lib.SpinnakerException as e:
+                    if e.errorcode == -1011:
+                        # The timeout was triggered. Nothing to worry about.
+                        continue
+                    elif e.errorcode == -1010:
```
jamienoss, line 290: [revision] Let's make an enum/macro for these error codes, please.

catkit/hardware/flir/flir_camera.py
```diff
@@ -0,0 +1,356 @@
...
+        # try-finally shuts down the camera if the generator is stopped.
+        try:
+            self.cam.BeginAcquisition()
+            frame_count = 0
+
+            # Get the pixel format to interpret the raw data.
+            if self.pixel_format == 'mono8':
+                pixel_format = self.instrument_lib.PixelFormat_Mono8
+            else:
+                pixel_format = self.instrument_lib.PixelFormat_Mono16
```
jamienoss, line 280: [revision] I could be wrong but this entire thing seems very circular and cries for an enum, you've implemented a similar mapping more than once now.

I would make an enum and then just assign a member to ``self.pixel_format`` and be done with it.

catkit/hardware/flir/flir_camera.py
jamienoss: Also, see https://github.com/spacetelescope/catkit/runs/4484824874?check_suite_focus=true#step:5:10

jamienoss (changes requested):
Thanks again for this, see revision comments.

catkit/hardware/flir/flir_camera.py
jamienoss: In case you read this and make alterations to un-name-mangle the "enum" before reading on, I would read on. I really get the impression that you just wanna use an ``Enum``.

catkit/hardware/flir/flir_camera.py
jamienoss: If the doc states that it doesn't raise, just add an inline comment, please.

catkit/hardware/flir/flir_camera.py
jamienoss: i.e., would ``_open()`` need to be re-called?

catkit/hardware/flir/flir_camera.py
ehpor: ~Also, no bare except (just adding a note to myself, I know it's part of your comment).~ nvm, I see your second comment now.

PR #244, opened 2022-02-09
Add code for McPherson 789A-4 scan controller

jamienoss:


jamienoss:
NOTE: Failing CI can be ignored, it's just that the upstream hicat testing that we do in catkit is temporarily broken.
Also note, that there's nothing in our test suite that tests any of this code anyhow. We can write the emulator for it if/when that  time comes (it requires a STUF simulator which I don't think is within the project requirments yet.)

catkit/hardware/mcpherson/scan_controller_789a4.py
```diff
@@ -0,0 +1,616 @@
...
+    READ_PARAMS = 'X'
+    GET_LIMIT_SWITCH_STATUS = ']'  # 0=no lim encountered, 32=home lim enc., 64=high lim enc., 128=low lim enc.
+    SCAN_UP = '+'
+    SCAN_DOWN = '-'
+    GET_MOTION_STATUS = '^'  # 0=not moving, 1=moving, 2=high in const vel, 16=slewing ramping complete.
+
+
+class LimitSwitchStatus(enum.Enum):
+    # NOTE: It's possible to be both HOME & LOW as the HOME region is wide enough to include the LOW limit.
+    NO_LIMIT = 0  # Not at any of the below. NOTE: Could be HOME but requires home switch to be enabled.
```
jamienoss: I don't like this naming, but couldn't think of anything better at the time, though perhaps ``INBETWEEN`` is better? 🤷

catkit/hardware/mcpherson/scan_controller_789a4.py
```diff
@@ -0,0 +1,616 @@
...
+        if is_home:
+            resp = resp - np.int32(LimitSwitchStatus.HOME.value)
+
+        status = LimitSwitchStatus(resp)
+
+        return status
+
+    def poll_limit_status(self, break_conditions, *args, **kwargs):
+        """ Poll device for limit status. See self.poll_status() for more info. """
+        return self.poll_status(break_conditions, *args, **kwargs)
```
jamienoss: This isn't used and could be deleted. However, if kept, it's currently wrong and should be:
```python
return self.poll_status(break_conditions, self.getlimit_status, *args, **kwargs)
```

catkit/hardware/mcpherson/scan_controller_789a4.py
```diff
@@ -0,0 +1,616 @@
...
+
+        :return list of read data.
+        """
+
+        data = []
+        try:
+            while True:
+                data.append(self.instrument.read_bytes(1))
+        except pyvisa.VisaIOError:
+            pass
```
jamienoss, line 313: [revision] These two funcs should ``return data``.

jamienoss: They should also be moved to a base class so they can be inherited both from this controller and the device controller in #247.

catkit/hardware/mcpherson/scan_controller_789a4.py
```diff
@@ -0,0 +1,616 @@
...
+        counter = 0
+        while counter < timeout:
+            status = func()
+            if status in break_states:
+                break
+
+            # NOTE: There's no need to sleep between iterations as there is already a query delay effectively doing the
+            # same thing.
+            if timeout is not None and timeout > 0:
+                counter += self.QUERY_DELAY
```
jamienoss, line 358: [revision] This func should actually time each loop rather than assuming each query duration is as expected.

jamienoss: This should also be pulled out so that it can be reused by the 747 device controller in #247. It can go in ``catkit.utils``.

jamienoss: Done in #247 and will propagate to here post-merge.

catkit/hardware/mcpherson/scan_controller_789a4.py
jamienoss: Done in #247 and will propagate to here post-merge.

aidancgray (approved):

PR #250, opened 2022-02-15
CATKIT-101 Add Service functionality

jamienoss:

### DIFF
--- a/catkit/interfaces/Instrument.py
+++ b/catkit/interfaces/Instrument.py
@@ -170,6 +170,8 @@ class Instrument(Service, MutexedNamespace, ABC):
         if not_permitted:
             raise TypeError(_not_permitted_error)
 
+        super().__init__(disable=not run_as_service)
+
         self.run_as_service = run_as_service
         self._context_counter = 0  # Used to count __enter__ & __exit__ paired usage.
         self.log = logging.getLogger()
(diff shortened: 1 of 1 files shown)


jamienoss:
@ehpor This is the start of the template we talked about the other day with @RemiSoummer.

We can work on it together, etc, from here.

catkit/interfaces/Instrument.py
```diff
@@ -96,10 +166,11 @@ class Instrument(MutexedNamespace, ABC):
     # Is an issue otherwise if an  __init__() raises before self.instrument = None is set.
     instrument = None
 
-    def __init__(self, config_id, *not_permitted, **kwargs):
+    def __init__(self, config_id, *not_permitted, run_as_service=False, **kwargs):
         if not_permitted:
             raise TypeError(_not_permitted_error)
```
jamienoss, line 169: [revision] Call ``super().__init__(disable = not run_as_service)``.

catkit/interfaces/Instrument.py
```diff
@@ -167,9 +241,20 @@ def _forced_safe_close(self):
...
+
             if self.instrument:
-                self._close()
-                self.log.info("Safely closed connection to '{}'".format(self.config_id))
+                try:
+                    self._close()
+                    self.log.info("Safely closed connection to '{}'".format(self.config_id))
+                except Exception:
+                    if not self.run_as_service:
+                        raise
```
jamienoss: I'll have to remember why I added this... 🤔

jamienoss: Right, the re-open below. Originally, ``service_thread`` was daemonic, but I no longer think that's a good idea.

PR #263, opened 2022-03-07
Refactor starlight xpress filter wheel to use pyusb

jamienoss:
- [x] Tested on hardware.


jamienoss:
``idVendor=0x1278, idProduct=0x0920``

PR #268, opened 2022-04-01
Fix issues with Standa stages

jamienoss:
Signed-off-by: The Mule <dummy@gmail.com>

### DIFF
--- a/catkit/hardware/standa/stages.py
+++ b/catkit/hardware/standa/stages.py
@@ -6,6 +6,7 @@
 # Generic Stage Class
 
 import ctypes
+import enum
 import math
 import os
 import platform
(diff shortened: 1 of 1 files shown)


catkit/hardware/standa/stages.py
```diff
@@ -91,11 +98,19 @@ def home(self):
         hmst.uHomeDelta = int(self.u_homeOffset)
```
jamienoss, line 98: @aidancgray This looked super suspicious so I tested it to make sure they weren't doing anything 🤯  and it doesn't work as suspected. I've added ``set_home_settings``.

catkit/hardware/standa/stages.py
```diff
@@ -246,44 +268,48 @@ def get_position(self):
...
         if result != self.instrument_lib.Result.Ok:
             raise RuntimeError("Soft stop failed")
 
-    def await_stop(self, *args, **kwargs):
+    def await_stop(self,  timeout=10*60, poll_interval=1):
         """ Wait for device to indicate it has stopped moving.
 
             See catkit.util.poll_status for API.
+
+            NOTE: Default timeout is arbitrarily long.
```
jamienoss, line 283: TODO: We could either make sure this is really long enough or calc an estimate based on speed.

PR #269, opened 2022-04-29
Change port for shared testbed state manager to 7001 (bug on MacOS Monterey)

ivalaginja:
@ehpor on you to describe what we found 🙏

### DIFF
--- a/catkit/multiprocessing.py
+++ b/catkit/multiprocessing.py
@@ -859,5 +859,5 @@ SharedMemoryManager.register("EventProxy", proxytype=EventProxy, create_method=F
 
 
 # Types used for CI testing.
-SHARED_STATE_ADDRESS = ("127.0.0.1", 7000)
+SHARED_STATE_ADDRESS = ("127.0.0.1", 7001)
 SharedState = MutexedNamespaceSingleton.factory(address=SHARED_STATE_ADDRESS, name="SharedState", timeout=2)


ehpor:
Port 7000, which is our default port for the shared testbed state manager, is used for a new AirPlay Receiver option in the new Monterey MacOS version. This feature is on by default, but can be turned off. As port 7000 is already opened by the AirPlay service, catkit will try to connect to that port for getting the testbed state. Apple AirPlay is not aware that it is supposed to be hosting our testbed state, and so it will never respond to our connection request, and the `manager.connect()` will hang indefinitely.

See https://developer.apple.com/forums/thread/682332 for more details (and outrage from other developers). Port 5000 & 7000 are standard ports for webdevelopers.

The fix is to use another default standard port. We chose 7001.

ivalaginja:
I can confirm that I can now run hicat-package without any issues on my MacOS Monterey.

jamienoss (dismissed):
@ehpor check the downstream hicat tests as I think some may use 7001.

ivalaginja:
The env file requires Python 3.7.6:
https://github.com/spacetelescope/catkit/blob/130c1f8ac2136071e465c7286b3df5e0da4efac7/environment.yml#L24

however the CI installs Python 3.7. which defaults to 3.7.13 now, which breaks the import of `Convolution` from `astropy`. I suppose the easiest would be to force the CI to install 3.7.6 as well? The version spec for the catkit tests is specified here:
https://github.com/spacetelescope/catkit/blob/130c1f8ac2136071e465c7286b3df5e0da4efac7/.github/workflows/CI.yml#L70

But there are other ones in that file, for the hicat tests and the linters. Shall we change all of them, to keep it consistent?

ivalaginja:
I tried doing the above in #273 but that didn't work.

jamienoss:
@ehpor @RemiSoummer The current CI is failing as you'll need to set up new secrets from @makidonlab.
See https://docs.github.com/en/actions/security-guides/encrypted-secrets and https://github.com/spacetelescope/catkit/blob/develop/.github/workflows/CI.yml#L121

RemiSoummer:
Thanks @jamienoss this issue appeared just yesterday, and was starting to look at the "secrets" things.  This must have been caused by the account transfer, Thanks for the links

RemiSoummer:
This has been tested successfully on hardware in addition to sims.

ehpor (approved):
Tested on hardware on Windows and on sims on Windows&MacOS.

PR #270, opened 2022-05-02
remove codeowners file

RemiSoummer:
@jamienoss and @kjbrooks have left the institute and are still overall code-owners for the entire repo and iris-ao respectively. 

This PR is removing the code-owners file to avoid any delays because of that.


RemiSoummer:
obviously this PR is waiting review from @jamienoss so I'm force merging it :-)

PR #271, opened 2022-05-02
Fixate python version in CI

ivalaginja:
Testing whether it works if I fix the used Python version in the CI to v3.7.6, as described in https://github.com/spacetelescope/catkit/pull/269#issuecomment-1114883367


ivalaginja:
whoops, forgot the base is not a protected branch, so it won't trigger the CI

PR #273, opened 2022-05-02
Fix failing CI caused by `photutils` importing astropy, `ImportError` with `astropy.convolve`

ivalaginja:
Testing whether it works if I fix the used Python version in the CI to v3.7.6, as described in https://github.com/spacetelescope/catkit/pull/269#issuecomment-1114883367

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -68,7 +68,7 @@ jobs:
       max-parallel: 1
       matrix:
         python-version: [3.7.9]
-        os: [ubuntu-latest, windows-2019]
+        os: [ubuntu-latest, windows-latest]
     steps:
     - uses: actions/checkout@v1
       with:
@@ -101,7 +101,7 @@ jobs:
       max-parallel: 1
       matrix:
         python-version: [3.7.9]
-        os: [ubuntu-latest, windows-2019]
+        os: [ubuntu-latest, windows-latest]
     steps:
     - uses: actions/checkout@v1
       with:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       max-parallel: 1
       matrix:
-        python-version: [3.7.6]
+        python-version: [3.7.11]
     env:
         # http://flake8.pycqa.org/en/latest/user/error-codes.html
         EXCLUDE_DIRS: catkit/hardware/newport/lib,catkit/hardware/boston/sdk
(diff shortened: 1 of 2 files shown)


ivalaginja:
Nope, it doesn't like it being specified like this.

ivalaginja:
Summary:

@zanecodes updated `setup-python` in GitHub actions to use v2, which allows us to specify minor Python versions. Fixing it to 3.7.6 did not help.
Fixing it to 3.7.11 but the CI couldn't install it.
In the most recent merged PR on catkit that successfully ran CI, the Windows tests used Python 3.7.9, so we tried this as well, which also failed.

We approached @pllim since this seems to be an astropy issue:
"astropy has dropped Python 3.7 support since v5.0, so you are stuck with an older and unsupported version of astropy. We no longer deploy fixes to the v4 series. Please use Python 3.8 or later."

We are using astropy 4.3.1 in the CI though, so they added:
"My hunch is that the CI provided C compiler got upgraded to an incompatible version or something. It is the C extension that failed to build."
and
"maybe change windows-latest to windows-2019 and see"
which is what I am trying right now.

Just giving it Python 3.7 again gives me a completely new error though, so now I am trying the combo of python 3.7.9 with windows-2019, waiting for the results.
Upgrading the Python version we use is not possible as we have dependencies on hardware libraries that are incompatible with newer Python versions than v.3.7

pllim:
Probably not related but `setup-python` is in v3 now.

ivalaginja:
Still the same error :(

ivalaginja:
Latest successful CI run on catkit:
https://github.com/spacetelescope/catkit/runs/5488240081?check_suite_focus=true

ivalaginja:
@RemiSoummer and I ran the failing test on the HiCATdeux machine (`catkit/tests/test_all_imports.py`) and this particular failing test passes, but a different one fails 💀

RemiSoummer:
This is what we get running the test_all_import test on Catkit using the existing hicat-package env on the hicat windows machine:   

image

ivalaginja:
Removing all `photutils` imports fixes the tests as any photutils import also imports `astropy` and that throws an `ImportError` for `astropy.convolve`.
Downstream hicat-package tests still import photutils so they will fail, however, this PR is now ready to be cleaned up and merged to have a new catkit release that fixes this.
@ehpor @RemiSoummer

RemiSoummer:
This PR is passing all the catkit tests, we are removing here all the dependencies to photutils from CATKit that should not have been there in the first place.   

HiCAT tests are failing, as expected, so I'm going to force merge that.

PR #274, opened 2022-05-02
test commit

RemiSoummer:
simple branch  to try to understand what's going on with the Astropy convolution problem


ivalaginja:
Not needed anymore, all work in #273

ivalaginja:
We had a new idea, so reponening

PR #275, opened 2022-05-03
added function to allow irisao command to be plotted in a subplot of …

sfr348:
https://jira.stsci.edu/browse/CATKIT-103

currently it is very difficult to plot an irisao command, this function facilitates that.  it does posses some re-used code from `display()`: so maybe we can make that cleaner by having `display()` call `get_wavefront_custom()`?

### DIFF
--- a/catkit/hardware/iris_ao/segmented_dm_command.py
+++ b/catkit/hardware/iris_ao/segmented_dm_command.py
@@ -417,6 +417,39 @@ class SegmentedDmCommand(object):
             plt.close()
         else:
             plt.show()
+    
+    def get_wavefront_custom(self):#,figure_name_prefix, out_dir, vmax=0.5e-6*u.meter, save_figure=True):
+        """
+        Plot the deployed mirror state (wavefront error)
+
+        :param figure_name_prefix: str, String to be added to filenames of output figures
+        :param out_dir: str, name of output directory
+        :param vmax: astropy unit quantity, the maximum value to display in the plot (as expected
+                     by Poppy)
+        :param save_figure: bool, If true, save out the figures in the directory specified by
+                            out_dir
+        """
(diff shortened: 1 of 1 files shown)


ivalaginja:
Hey @sfr348, could you precise what this new function's intention is and how it differs from `display()`? It seems like it is doing exactly the same thing but without the actual plotting, and by getting rid of all plotting defaults. Is that the intended behavior? What is your use case?

ivalaginja:
Also, be mindful that you created this branch out of the tagged version, not develop. This shouldn't matter much in your case because the branch will need to be rebased on dev anyway after we fix the CI issue, but something to keep in mind for the future, it would save you some work.

sfr348:
>  What is your use case?

So I wanted to be able to plot the irisAO command as a subplot in a figure post-experiment (like for making talk slides etc); I provided the sample code to do this in the description of `get_wavefront_custom()`.  I could use `display()` to update the opd and just produce a throwaway plot but that seemed annoying.  I just updated it to remove the duplicate code, I didn't want to mess with any of the base functions initially when I didn't have my own branch.  I can also change the name of `get_wavefront_custom()` to something more intuitive

ivalaginja:
@sfr348 this looks good to me, however the name of the new function could be clearer, as you indicate yourself. Looking at what the new function does, it retrieves the current data of the segmented command in question and applies it to the `aperture`, which is a simulated proxy we use for plotting. So maybe you could rename the function into something like `apply_current_data()` or `apply_current_wavefront()` or something the like?

Apart from that, you'll have to rebase this branch after we merged some things recently into dev.

sfr348:
@ivalaginja all rebased against develop.

ivalaginja (approved):
good by me

PR #277, opened 2022-05-09
Remove sign flip depending on Poppy version.

ehpor:
This is now handled by the hicat simulator. See https://github.com/spacetelescope/hicat-package/pull/628.


RemiSoummer (approved):
I agree this fixes the issue, and this this has no impact on version 0.92

